PR #7, opened 2015-04-24
Note about starting fig-elk when SELinux is enabled

cdituri:
Centos|Redhat is used in the example since that was the type of distribution I was on. Regardless, the note should point future readers, who might be in the same predicament, in the right direction.


deviantony:
Thanks @cdituri !

cdituri:
no problem, anytime ;-)

PR #13, opened 2015-09-14
Add redis queue to elk stack

rgroten:
Added a redis container to act as message queue

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@ elasticsearch:
     - "9200:9200"
 logstash:
   image: logstash:latest
-  command: logstash -f /etc/logstash/conf.d/logstash.conf
+  command: logstash -f /etc/logstash/conf.d
   volumes:
     - ./logstash/config:/etc/logstash/conf.d
   ports:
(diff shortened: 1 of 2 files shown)


deviantony:
Hello, thanks for the contribution but I'm not sure that the default stack should be using a more complex setup than using only the ELK components.

I could ponder a way using branches to expose multiple working setups and document it in the README.

I'll keep the PR open till I found a solution.

rgroten:
No problem, I needed it for my solution and thought I'd share it in case you're interested!  Your project was extremely useful btw.

deviantony:
Thanks :+1: 

As I said, this could be interesting to have people sharing their working configurations on another branch/something, I need to think about this a little.

deviantony:
I'll close this PR, your work should be referenced somewhere else.

mskutin:
@deviantony why you won't just merge this into a separate branch? 

> I could ponder a way using branches to expose multiple working setups and document it in the README. 

 I think it's a good idea to keep multiple scenarios in one repo, you can split the `default` elk stack, keep it in main branches and create subdirectories for the Vagrant, Redis, etc.

deviantony:
The discussion about branches/wiki/files is opened in #33. Not decision at the moment.

PR #29, opened 2015-12-12
kibana is now executed using exec

fvigotti:
kibana should be executed with exec to give it pid 1 and bind signal forwarding with the container instance


deviantony:
Yeah thanks !!

PR #38, opened 2016-02-03
Make ELK containers always restart.

wizardbyron:
add `restart: always` to each one.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,10 +1,13 @@
 elasticsearch:
+  restart: always
   image: elasticsearch:latest
   command: elasticsearch -Des.network.host=0.0.0.0
   ports:
     - "9200:9200"
     - "9300:9300"
+
 logstash:
+  restart: always
   image: logstash:latest
   command: logstash -f /etc/logstash/conf.d/logstash.conf
   volumes:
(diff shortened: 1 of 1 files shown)


deviantony:
What is the advantage of adding this configuration to the containers?

From the official doc:

Always restart the container regardless of the exit status. When you specify always, the Docker daemon will try to restart the container indefinitely. The container will also always start on daemon startup, regardless of the current state of the container.

Personally, I do not want the stack to boot up every time my docker-engine starts (OS startup).

kylegordon:
What's your thoughts on using ```restart: unless-stopped``` instead?

antoineco:
The question is: do _most_ people want their containers do be restarted behind their back?

I don't think so. Docker doesn't restart containers by default, neither should we. People expect a different behaviour can easily set the `restart` parameter, but forcing it on _all_ users doesn't sound reasonable.

kylegordon:
Yes, that's why I'm asking about using the 'unless-stopped' value.

If you have a running docker instance, I would expect it to be running again after the server is rebooted. If I have manually stopped it (or, indeed, never started it), then of course, I would not expect it to be running after a reboot.

antoineco:
I got it, and it may be indeed desirable for you, but again, the default behaviour of docker is to not restart stopped containers.
I'm not questioning whether the setting is safe to use, just trying to keep a consistent (but configurable) behaviour here.

PR #42, opened 2016-02-23
Fix error "[9] System error: not a directory"

minhdanh:
When starting the stack `with docker-compose up`, it gives out the error:

```
ERROR: Cannot start container e032695b38cdcd1c72597004f2369a7127b8233c9ecc2320c3b2bbd10209654b: [9] System error: not a directory
```


deviantony:
Thanks @minhdanh !

:+1:

PR #48, opened 2016-03-23
Convert docker-compose to version 2

mamoit:
Addresses issue #45.
README.md needs some corrections.

Issue #45, opened 2016-03-15
Upgrade to compose file format version 2


deviantony:
There is a new version of the compose file format, the current will be deprecated in the future.

The new file fomat version requires the following:

* Docker engine 1.10.x+
* Compose 1.6.x+

See: https://docs.docker.com/compose/compose-file/#versioning

The format will not be updated at the moment, as I'm not sure about the tooling versions of the user base.

MBuffenoir:
Hey, maybe this can help: I have a v2 version here: https://github.com/skippbox/docker-on-cluster-howtos/blob/master/swarm/apps/elk/docker-compose.yml compatible with swarm (i.e no links)

deviantony:
Sweet, thanks ! I'll have a look at it when the time will come to upgrade to the new version of the compose format.

mamoit:
:+1:

deviantony:
Thanks for the PR @mamoit ! :+1: 

Has I said before, I'm not sure when I should update the compose file to the new format.

mamoit:
No need to accept it just for the sake of the upgrade (no need to accept at all for that matter).
I converted the docker-compose file just to explore elk, so I decided to share what I did.

HackerWilson:
Hey, guys. i wrote the v2 version there: https://github.com/HackerWilson/docker-elk-deployment

with redis broker and logstash shipper support.

i am not sure that if i change the readme to english, then can it be your version2?

deviantony:
@HackerWilson I'd rather not change the structure of the project by keeping the actual number of components (so no redis/logstash shipper).

Anybody know what upgrading to the new format could bring to this project (beside solving a future deprecation)?

mamoit:
Does the old format support network overlays in a swarm cluster?
I think the main bonuses would be facing towards swarm support and integration.
But again, I'm still exploring, so I can't say I'm sure about anything right now.

HackerWilson:
@deviantony I am new to elk, and I was wondering that how many logs can elk deal with when deployed without redis/logstash shipper support?

deviantony:
@mamoit Never played with swarm, I got to check it though.

@HackerWilson Well I did use a production setup with a multi node Elasticsearch cluster and 1 logstash indexer node, it was able to deal with 30M events/day.

HackerWilson:
@deviantony hmn, good.

MBuffenoir:
Hi,

[...]

PR #54, opened 2016-05-17
Docker-compose.yml updated to version2.

norbertkorodi:
Solved kibana network issue. The linking for kibana->elastic was not set (seems most likely an internal docker bug to me), but upgrading to version 2 automatically solved the issue.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,23 +1,25 @@
-elasticsearch:
-  image: elasticsearch:latest
-  command: elasticsearch -Des.network.host=0.0.0.0
-  ports:
-    - "9200:9200"
-    - "9300:9300"
-logstash:
-  image: logstash:latest
-  command: logstash -f /etc/logstash/conf.d/logstash.conf
-  volumes:
-    - ./logstash/config:/etc/logstash/conf.d
-  ports:
-    - "5000:5000"
-  links:
-    - elasticsearch
-kibana:
-  build: kibana/
-  volumes:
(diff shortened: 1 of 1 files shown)


deviantony:
Already open in #48 

See the discussion in #45 on why I do not accept PR on compose file V2 for now.

deviantony:
I'll keep it open as the content of this PR differs from the one in #48. I'll need to figure out the better one.

norbertkorodi:
Alrighty.

gabel:
Version 2 will enable users to use build args which is required in certain enterprise proxy environments.... 😢 

> ## args
> Version 2 file format only.

https://docs.docker.com/compose/compose-file/#/args

deviantony:
@gabel I'm thinking to switch to v2 for the `x-pack` and `5.x` branch first.

gabel:
@deviantony no hurry at all - it's not that hard to change the format in an own fork.

deviantony:
Closed ! `master` is now using compose v2 format. Thanks for the help !

gabel:
@deviantony Thanks... 

With extends (https://docs.docker.com/compose/extends/) it is now easy to add proxy environments.


Example `docker-compose.override.yml`
```yml
version: '2'

services:
  elasticsearch:
    build:
      context: elasticsearch/
      args:
        - http_proxy
        - https_proxy
        - no_proxy
    environment:
      - http_proxy
      - https_proxy
      - no_proxy
```

PR #55, opened 2016-06-13
Feature: allow logstash plugin management

michaeltarleton:
I wasn't able to figure out another way to install logstash plugins so I created a docker build script that allows for adding plugins.

I needed to add the logstash-filter-json plugin to parse out my "message" but I could see where having this as a default for everyone would not work (ie. if they save xml in their message)

### DIFF
--- a/logstash/config/logstash.conf
+++ b/logstash/config/logstash.conf
@@ -4,7 +4,7 @@ input {
 	}
 }
 
-## Add your filters here
+## Add your filters / logstash plugins configuration here
 
 output {
 	elasticsearch {
--- a/logstash/config/logstash.conf
+++ b/logstash/config/logstash.conf
@@ -6,6 +6,13 @@ input {
 
 ## Add your filters here
 
+# Added a default json filter for parsing the "message" field
+filter {
+    json {
+        source => "message"
+    }
+}
+
 output {
 	elasticsearch {
 		hosts => "elasticsearch:9200"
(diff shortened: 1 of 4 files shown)


deviantony:
Hey ! Thanks for th PR !

Two things here:

* I really like the section in the README about how to add logstash plugins.
* I do not think that it should ship the `logstash-filter-json` as its outside of the scope of this project, which is providing a basic configuration.

Maybe we should just add a empty Dockerfile for logstash in `logstash/Dockerfile` with the following declaration:

```
FROM logstash:latest

# Add your logstash plugins setup here
# Example: RUN logstash-plugin install logstash-filter-json 
``` 

Also, could you update the README section with the following:

*Add the associated plugin ~~code~~ configuration to the `logstash/config/logstash.conf` file*

michaeltarleton:
I see what you mean. I added two more commits in response to your comments.

logstash/config/logstash.conf
```diff
@@ -6,13 +6,6 @@ input {
 
 ## Add your filters here
```
deviantony: Replace this with:

```
## Add your filters / logstash plugin configuration here
```

deviantony:
Nice, just check out my comment and we're all good !

michaeltarleton:
I think we have a winner!

deviantony:
Perfect ! Thanks man !

michaeltarleton:
You're welcome. Thank you for starting this repo. It's the best dockerized ELK stack I've found so far!

PR #67, opened 2016-09-26
Upgrade to version 2 of the Docker Compose format

kurron:
I've modified the compose file to use the new version 2 format.  I've tested it and the containers still come up and function as expected.

Thanks,
Ron

ps. 

Thanks for setting up the ELK stack.  Saved me having to figure everything out on my own.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,24 +1,28 @@
-elasticsearch:
-  image: elasticsearch:latest
-  command: elasticsearch -Des.network.host=0.0.0.0
-  ports:
-    - "9200:9200"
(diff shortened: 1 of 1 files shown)


deviantony:
Yeaye another v2 format PR :sunglasses: 

Similar to #54 & #48

As said in #54, I'll switch to v2 format when x-pack will be released. The `master` branch will then point to an ELK 5 x-pack setup in Docker Compose V2 format.

Thanks again ! And you're welcome :)

deviantony:
Closed ! `master` is now using compose v2 format. Thanks for the help !

PR #73, opened 2016-10-12
added Vagrantfile to run the stack

andmos:
I found this project really useful, thanks for making it available! 
I like to use Vagrant as a part of the Hashicorp-flow when deploying stuff, so I created a Vagrantfile for running the project automaticly. It might be usefull for more people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,28 @@ So you have to either
 
 See [issue 36](../../issues/36) for details.
 
+
+
+## Vagrant 
+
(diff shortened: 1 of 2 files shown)


deviantony:
Branches are not really documented on this project, but we do already have a Vagrant based stack: https://github.com/deviantony/docker-elk/tree/vagrant

Thanks ;)

andmos:
Ah did not catch that. Great work!

PR #78, opened 2016-11-03
Use depends_on dependency container

nathaliaortegaGL:
add depends_on to kibana and logstash on docker-compose.yml to prevent having to install NC and not have to keep checking for ES to start as the container will start after.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,6 +19,8 @@ services:
       - "5000:5000"
     networks:
       - docker_elk
+    depends_on:
+      - elasticsearch
   kibana:
     build: kibana/
     volumes:
@@ -27,6 +29,8 @@ services:
       - "5601:5601"
     networks:
       - docker_elk
+    depends_on:
+      - elasticsearch
 
 networks:
   docker_elk:
--- a/kibana/Dockerfile
+++ b/kibana/Dockerfile
@@ -1,7 +1,5 @@
 FROM kibana:5
 
-# RUN apt-get update && RUN apt-get install -y netcat bzip2
-
 COPY entrypoint.sh /tmp/entrypoint.sh
 RUN chmod +x /tmp/entrypoint.sh
 
--- a/kibana/Dockerfile
+++ b/kibana/Dockerfile
@@ -1,3 +1,1 @@
 FROM kibana:5
-
-CMD ["/tmp/entrypoint.sh"]
--- a/kibana/Dockerfile
+++ b/kibana/Dockerfile
@@ -1,6 +1,3 @@
 FROM kibana:5
 
-COPY entrypoint.sh /tmp/entrypoint.sh
-RUN chmod +x /tmp/entrypoint.sh
-
 CMD ["/tmp/entrypoint.sh"]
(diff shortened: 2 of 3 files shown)


deviantony:
Oh this looks really nice !! Did you try it? Are you sure that the kibana service will be able to detect that the elasticsearch is ready to listen on port 9200 ?

As stated in docker-compose docs: https://docs.docker.com/compose/compose-file/#/dependson

> Note: depends_on will not wait for db and redis to be “ready” before starting web - only until they have been started.

Maybe Kibana 5 now have a better support to re-establish connections to elasticsearch though... I need to try this :)

quentinfayet:
@deviantony I'm currently using `depends_on` on a custom docker ELK stack, and everything works just fine. Kibana service waits quietly for Elasticsearch to be up and running.

nathaliaortegaGL:
@deviantony yes, I tried it and it works fine! :)

kibana/Dockerfile
```diff
@@ -1,7 +1,5 @@
 FROM kibana:5
 
-RUN apt-get update && apt-get install -y netcat bzip2
-
 COPY entrypoint.sh /tmp/entrypoint.sh
```
deviantony: You can also remove the lines 3-4 which are used to copy the `entrypoint.sh` script. This script has no purposes anymore, so please delete it too.

kibana/entrypoint.sh
```diff
@@ -1,10 +1,4 @@
 #!/usr/bin/env bash
 
-# Wait for the Elasticsearch container to be ready before starting Kibana.
```
deviantony: As said previously, delete the file `kibana/entrypoint.sh`.

docker-compose.yml
```diff
@@ -19,6 +19,8 @@ services:
       - "5000:5000"
     networks:
       - docker_elk
+    depends_on:
+      - elasticsearch
```
deviantony: I've just tested, the whole stack actually works even without the `depends_on` directive :) For a matter of clarity, we'll keep these directive though.

deviantony:
ping @nathaliaortegaGL

krainboltgreene:
depends_on will (1.10) eventually use HEALTHCHECK to determine if it's actually ready, which is nice, but any "it works in correct order" here is chance.

deviantony:
Yeap, this is indeed a good PR. Another ping @nathaliaortegaGL ?

ernestova:
I just ping @nathaliaortegaGL internally  to update her PR.

nathaliaortegaGL:
@deviantony changes have been applied!

kibana/Dockerfile
```diff
@@ -1,8 +1,3 @@
 FROM kibana:5
 
-RUN apt-get update && apt-get install -y netcat bzip2
-
-COPY entrypoint.sh /tmp/entrypoint.sh
-RUN chmod +x /tmp/entrypoint.sh
-
 CMD ["/tmp/entrypoint.sh"]
```
deviantony: You also need to remove this line, otherwise it won't work.

deviantony:
@nathaliaortegaGL one more thing needs to be updated.

nathaliaortegaGL:
@deviantony Sorry, forgot to delete that :) It's ready now.

deviantony:
Thanks @nathaliaortegaGL :)

PR #95, opened 2017-02-27
Use new official images (X-Pack)

antoineco:
*First of all thanks for sharing this setup! It's lean and uses sane defaults, perfect for prototyping.* :+1:

---

The images from the Docker Hub are now deprecated in favour of new images maintained by Elastic. 

Notable changes:
- **These images include X-Pack by default**, therefore there is no need to install it explicitly
- The Logstash configuration is split between `config/` (holds the agent configuration) and `pipeline/` (holds the input/output plugins configuration)

---

I would gladly backport this PR to the other branches as well but I would like to know your opinion first regarding the way to proceed in order to disable authentication. Would you rather:

- enable [anonymous auth](https://www.elastic.co/guide/en/x-pack/current/anonymous-access.html)?
- disable it using [configuration flags](https://www.elastic.co/guide/en/x-pack/current/installing-xpack.html#xpack-enabling)?
- [uninstall X-Pack](https://www.elastic.co/guide/en/x-pack/current/installing-xpack.html#xpack-uninstalling) in all images?
- something else?


deviantony:
Hi @antoineco ! Thanks for the words and for the work !

I'd accept a PR that disable X-Pack using configuration flags for the master branch, I think that should be good enough and would keep the setup simple. 

I did a quick review of your PR, nice work, it looks even simpler now :)

antoineco:
Alright, thanks for your answer, let me send a separate PR.

PR #104, opened 2017-03-27
Update Searchguard to 5.2.2

salimhb:
I wasn't able to start it from the searchguard branch due to errors saying that the search-guard plugin is not compatible with the installed elasticsearch version.


deviantony:
Thanks @salimhb !

PR #110, opened 2017-04-13
Add Logspout functionality for out-of-the-box Docker logging

joaquincasares:
This PR includes the following changes:

* Adds the Logspout container which captures all Docker Logs and sends them to Logstash.
* Capitalizes all Logstash references in the README.
* Uses the term Docker Compose, instead of Docker-compose, in the README.
* Auto-creates the Logspout index.
* Introduces the [.env file](https://docs.docker.com/compose/env-file/) to be used for auto-creating indexes, and soon for simplified ELK Stack upgrades.
* Made a questionable change to the Logstash pipeline config. I know this works for Logspout, but will it work for all other use cases?

Great work on this project! Definitely saved me a great deal of time!


antoineco:
@joaquincasares great job! I think this PR could benefit to some people, thanks for submitting it :+1: 

A few comments on my side, which other reviewers may add to if they feel like it:

---

#### 1. Not everyone wants Logspout to suddenly start logging everything running on their local Docker

Therefore I'd move the Logsput part to a separate docker-compose file, that way users can start the stack with eg. `-f logspout.yml` if they want Logspout to be included.

#### 2. The use of `.env` is incorrect in this context

From the Compose [documentation](https://docs.docker.com/compose/compose-file/#envfile):
> Note: If your service specifies a build option, variables defined in environment files will not be automatically visible during the build. Use the args sub-option of build to define build-time environment variables.

Also the way you want to use it will be available only in very recent Docker versions, we want to also support slighly older builds. Besides you're using this file only for Kibana, which makes it incomplete.

#### 3. Some bits may be too opinionated

I'm not a big fan of overriding the Kibana startup script completely just for creating Logstash indexe pattern for 2 reasons. First, it forces us to keep it in sync manually with the upstream `kibana-docker` repo, suboptimal. Second, what if users want to have control over the creation of the index pattern?

You also added a `udp` input but I can not see where it's being used.

---

I'd be curious to read @deviantony's take on that

joaquincasares:
Thanks for the feedback @antoineco !

1. I've split up the docker-compose.yml, created a logspout.yml, and updated the README with instructions on how to include Logspout.

2. Those are fair points. I wanted to be future-looking, but you're correct that we should support older Docker environments. I've removed the .env file and related entries.

3. I've removed the auto-creating index script and updated the README with those commands. I've also removed the `udp` input section. Tbh, I don't actually know why it was included in the example that I referenced, but if you think it's best to remove it, I've done so. :)

Thanks again for the feedback! Hopefully the PR looks better now.

antoineco:
Thanks, I'll review the changes again later but at first glance it looks good.

Regarding filebeat, how about writing a wiki page that describes how to I use it with this project and reference the relevant official documentation pages when they're relevant?

joaquincasares:
Thanks for the follow up review!

I'd much rather use Dockerfiles as code/documentation, just since it's cleaner and by using tagging we'll know specifically when something worked, rather than have to make assumptions with stale documentation that may not have been updated in concert.

If creating a `./filebeat.yml` is too much, how about a `./filebeat/` directory which holds the Dockerfile, yml, and includes its own README?

Fwiw, I still think we should make `./filebeat.yml` a first-class citizen, like `./logspout.yml`, but I understand we may be adding too many blades to the Swiss Army Knife, which is why I figured to ask up front. :)

Personally, I have referenced this project many times since there's very little out there in regards to a "full-stack deployment" of all moving pieces, especially not one using official images. Now that I've started using the ELK stack more, I don't see a case where Filebeat would not at least be considered for a proper deployment.

But again, I'm down with helping toss some information into this repo, wherever it lands, in hopes of making it "THE" source for this sort of documentation. :)

antoineco:
Totally agree, but Filebeat is a log shipper, it's something which you deploy with your application, not on the ELK stack itself. Which is why I believe our responsibility ends here, and the rest is up to the user to chose how he wants to leverage the stack.

Filebeat is used widely out there, so maybe we could consider having an input for it in the default logstash conf, but including a Dockerfile makes sense only if you have a concrete application to send logs for, not to mention that every application will be configured differently (extra fields, multi line pattern, etc.).

What I mean is that I'd rather provide good guidance over a fragile customized integration which may not fit everyone's need. But maybe you already have a plan for that?

In any case the project owner will give his decision about whether we merge or not :)

logspout.yml
```diff
@@ -0,0 +1,18 @@
...
+services:
+  logspout:
+    build: logspout/
+    volumes:
+      # security concern:
+      # https://raesene.github.io/blog/2016/03/06/The-Dangers-Of-Docker.sock/
+      # http://stackoverflow.com/questions/40844197
+      - /var/run/docker.sock:/var/run/docker.sock:ro
+    environment:
+      ROUTE_URIS: "logstash+tcp://logstash:5000"
```
antoineco: What's the advantage of using this over `github.com/gliderlabs/logspout/transports/tcp`?

deviantony:
This seems to represents a good amount of work ! I'd like to see that as an extension rather than simply merging it inside the repository. As @antoineco said, this could benefit to some people, I just don't think all the users of this project.

We might just want to decide how to add 'extensions' inside this project, as this one brings quite a lot of changes, I don't think that a Wiki entry would be convenient. I like the ability to start some parts of the stacks using extra compose files, extensions would require to be fully automated if possible and have their own documentation section.

@joaquincasares @antoineco what do you think? suggestions on that?

logspout.yml
```diff
@@ -0,0 +1,18 @@
+version: '2'
+
+services:
+  logspout:
+    build: logspout/
+    volumes:
+      # security concern:
```
deviantony: You can probably remove that comment, it's more oriented towards global Docker usage than related to this project.

logspout/Dockerfile
```diff
@@ -0,0 +1,5 @@
+# uses ONBUILD instructions described here:
+# https://github.com/gliderlabs/logspout/tree/master/custom
+
+FROM gliderlabs/logspout:master
```
deviantony: What's the point of building your own logspout image?

logstash/pipeline/logstash.conf
```diff
@@ -1,7 +1,11 @@
+## based off: https://github.com/deviantony/docker-elk and
```
deviantony: You can remove these comments too.

logspout/modules.go
```diff
@@ -0,0 +1,9 @@
+package main
```
deviantony: Where is that file used?

logspout.yml
joaquincasares: From the best I can gather, tcp connections will retry sending dropped metrics.

I did find this: 

`retry tcp errors and don't hang forever on failure`

--
 https://github.com/gliderlabs/logspout/blob/c2bafe3ea23a2372cf4d72bfc06785d819ad35d7/CHANGELOG.md#v3---2016-03-03

logspout/Dockerfile
joaquincasares: I was following the instructions here to include the logspout to logstash adapter:

https://github.com/gliderlabs/logspout/tree/c2bafe3ea23a2372cf4d72bfc06785d819ad35d7/custom

logspout/modules.go
joaquincasares: It gets included with the ONBUILD commands:

https://github.com/gliderlabs/logspout/blob/c2bafe3ea23a2372cf4d72bfc06785d819ad35d7/Dockerfile#L9-L11

joaquincasares:
> not to mention that every application will be configured differently (extra fields, multi line pattern, etc.).

> I'd rather provide good guidance over a fragile customized integration which may not fit everyone's need.

@antoineco Those two are fair points. I wasn't going to do any sort of pattern matching and you're correct that that would lead to poor guidance. I'm down for doing something else for Filebeat.

@deviantony Glad your liking and thanks for starting up the project yourself! :) 

I've made the requested changes, but not sure I follow your idea on "extensions". Could you elaborate?

Also, when you mention plenty of changes, are we talking about Logspout or Filebeat? Sorry for doubling up on the thread! :)

> I like the ability to start some parts of the stacks using extra compose files

I'm +1 on that.

README.md
```diff
@@ -72,9 +82,29 @@ By default, the stack exposes the following ports:
 * 9300: Elasticsearch TCP transport
 * 5601: Kibana
 
-*WARNING*: If you're using *boot2docker*, you must access it via the *boot2docker* IP address instead of *localhost*.
+*WARNING*: If you're using `boot2docker`, you must access it via the `boot2docker` IP address instead of `localhost`.
+
+*WARNING*: If you're using *Docker Toolbox*, you must access it via the `docker-machine` IP address instead of `localhost`.
+
+# Create Indexes
```
antoineco: You might want to merge that with the line 77 (_"Refer to 'Connect Kibana with Elasticsearch' for detailed instructions about the index pattern configuration."_)

The wording is not 100% accurate, you're creating an index **pattern** here.

README.md
```diff
@@ -72,9 +82,29 @@ By default, the stack exposes the following ports:
...
+
+When Kibana launches, it will not contain any out-of-the-box indexes. Instead,
+you can run this command to create a Logstash index:
 
-*WARNING*: If you're using *Docker Toolbox*, you must access it via the *docker-machine* IP address instead of *localhost*.
+```bash
+$ curl -XPUT http://elasticsearch:9200/.kibana/index-pattern/logstash-* -d \
```
antoineco: Will not work from localhost because it can't resolve "elasticsearch". Maybe worth elaborating a little bit?

README.md
```diff
@@ -72,9 +82,29 @@ By default, the stack exposes the following ports:
...
+    '{"title" : "logstash-*"}'
+```
+
+This command will mark the Logspout index as the default Kibana index:
```
antoineco: Is this Logspout specific?

logspout/Dockerfile
antoineco: So the whole point is to include the "github.com/looplab/logspout-logstash" plugin, right?

logstash/pipeline/logstash.conf
```diff
@@ -1,7 +1,8 @@
 input {
-	tcp {
-		port => 5000
-	}
+    tcp {
+        port  => 5000
+        codec => json
```
antoineco: Not sure whether "json" is suitable for people who send raw lines, like in the `nc` example from the README.

deviantony:
@joaquincasares I'm thinking of extensions as modular parts that could be added quite easily to this project while still allowing to not use them. E.g. I'd like to be able to use the current ELK setup and later on maybe trigger the addition of logspout just by using another command (that would be the better user experience IMO) but I don't really have time to figure out how to implement that atm. Hence why I'm asking if either you or @antoineco have any ideas here ;)

logspout/Dockerfile
joaquincasares: Yes, that's correct.

README.md
joaquincasares: Thanks for the correction. Will update shortly.

README.md
joaquincasares: Good catch. Updating to include the `docker-compose run` command.

README.md
joaquincasares: Must have been my mistake. Correcting soon. Thanks for catching!

logstash/pipeline/logstash.conf
joaquincasares: I think that may be logspout specific. I'm having problems testing the current setup, but I'll ensure things run and see if we can do without that line.

joaquincasares:
Thanks for that clarification.

For "extensions", I'd go with @antoineco 's approach in bullet #1: https://github.com/deviantony/docker-elk/pull/110#issuecomment-294446515.

where we use something like: `$ docker-compose -f docker-compose.yml -f logspout.yml up` to add Logspout. Then we can continue making additional `compose.yml` files for each modular component.

dm:
Hi @antoineco @deviantony,

I'm diverging from this to a new canonical repo (changing away from official Elastic.co docker images) but wanted to get this working either way.

So before implementing incompatible changes, I've forked this, rebased and squashed these changes and will open a new PR. (Just going through local testing)

Cheers,
@dm

PR #121, opened 2017-05-10
disabled default password of xpack

kiranps:
if we need to completely disable xpack. we need to disable default password too

https://www.elastic.co/guide/en/x-pack/current/security-settings.html#password-security-settings


kiranps:
sorry

antoineco:
`xpack.security.enabled: false` so this is not necessary? Is this why you closed the PR?

kiranps:
elasticsearch.yml was not linking with /usr/share/elasticsearch/elasticsearch.yml due some unknow problem, so container was using default config which enabled xpack. i thought it was due some missing configuration

PR #140, opened 2017-06-30
added GELF support to the images and logstash.conf

er1z:
GELF is a common protocol, this PR enables support in Logstash container.


deviantony:
Hi @er1z and thanks for the contribution !

Unfortunately, I'm not going to merge this as we don't aim to support "common/most used" logstash protocols. This project has been created to help users bootstrap a basic ELK setup and I don't think that GELF support is required by all our users.

PR #144, opened 2017-07-01
plugin readme update

TomonoriSoejima:
I had to run "docker-compose build" after modifying Dockerfile.
Unless doing so, plugin was never added.

 And I thought adding this step will help others with limited knowledge of Docker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,7 @@ To add plugins to any ELK component you have to:
 
 1. Add a `RUN` statement to the corresponding `Dockerfile` (eg. `RUN logstash-plugin install logstash-filter-json`)
 2. Add the associated plugin code configuration to the service configuration (eg. Logstash input/output)
+3. Rebuild the images using the `docker-compose build` command
 
 ### How can I enable the provided extensions?
 


arubalcaba:
first of all I just want to say what an awesome job on this docker image. not sure if this is the right place to post this question but I'm trying to get the X-pack stuff going and I download the X-pack branch and ran it and everything comes up just fine except that Kibana never ask me login with a default user of any kind. I'm wondering if I have to enable that feature somehow before I startup the stack with docker-compose.  Any insight into this issue would be greatly appreciated.

TomonoriSoejima:
docker-elk/elasticsearch/config/elasticsearch.yml has

`xpack.security.enabled: false`

You have to change that to `true` and Kibana will ask you for password.

antoineco:
@arubalcaba it's not the best place indeed :) It would be better to open a new issue or ask your question on Gitter.

You don't have to change anything if you checked out the x-pack branch. If you run:

```
git checkout x-pack
docker-compose build
```

you will end up with the expected images, and see a login screen.

PR #151, opened 2017-07-14
docker-compose exec for searchguard initialized

saggiyogesh:
`docker-compose exec` makes more sense to initialize searchguard, as all the commands using `docker-compose`.


antoineco:
Fair point, you can also remove the `docker` command in that case.

PR #161, opened 2017-08-08
Searchguard

alexanderlc:


antoineco:
Opening and immediately closing PRs is getting trendy around here :)

PR #181, opened 2017-10-13
Adding new features

ivanrome:
Hello,

- Adding parsing apache logs
- Adding travis and circleCI support.
- Minor fixes


antoineco:
Could you clarify the intention here @ivanrome? Especially:
 * why both Travis and Circle?
 * what dependencies does DependencyCI check?
 * why should we add apache filters?

ivanrome:
Hello, 

- CircleCI and Travis are quite the similar. Actually I have both to see advantages in each one. At least one of them is a good to have.
- DependencyCI check is another test. I'd like to show docker images dependencies but for now I have not the answer.
- Apache filter is the special feature I want to add for profesional reasons.  I have apache logs that I'll need to extract some data to dimension the application

antoineco:
So if I understand well you opened that PR by mistake right? If that's the case can we close it?

_Note: I like the idea of having a CI testing our branches._

ivanrome:
Not a mistake, but maybe lot of CI integration & Apache logs you dont need.

Os Ok to me to close

antoineco:
Indeed I don't think the Apache integration is relevant here, but the CI integration is great! 👏 

Feel free to open for Travis as soon as you consider it's ready :)

PR #183, opened 2017-10-14
Travis integration

ivanrome:
Travis integration :gear:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,6 @@ sudo: required
 
 language: generic
 
-services:
-  - docker
-
 env:
   - DOCKER_COMPOSE_VERSION=1.14.0
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 sudo: required
 
-language: ruby
+language: generic
 
 services:
   - docker
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,37 @@
+sudo: required
+
+language: ruby
+
+services:
+  - docker
+
+env:
+  - DOCKER_COMPOSE_VERSION=1.14.0
+
+install:
+  # Installing docker
+  - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+  - sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+  - sudo apt-get update
+  - sudo apt-get -y install docker-ce
+
+  # Installing docker-compose
+  - sudo rm /usr/local/bin/docker-compose
+  - curl -L "https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-$(uname -s)-$(uname -m)" > docker-compose
+  - chmod +x docker-compose
+  - sudo mv docker-compose /usr/local/bin
+
+before_script:
+  # Check docker & docker-compose versions
+  - docker --version
+  - docker-compose --version
+
+script:
+  - docker-compose build
+  - docker-compose up -d
+
+after_success:
+  - sleep 4m
+  - docker network ls
+  - curl --retry 10 --retry-delay 5 -v http://localhost:9200/
+  - curl --retry 10 --retry-delay 5 -v http://localhost:5601/
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,6 @@ script:
 
   # Verifications
   - sleep 4m
-  - docker network ls
+  - docker-compose logs
   - curl --retry 10 --retry-delay 5 -v http://localhost:9200/
   - curl --retry 10 --retry-delay 5 -v http://localhost:5601/
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ script:
   - docker-compose build
   - docker-compose up -d
 
-after_success:
+  # Verifications
   - sleep 4m
   - docker network ls
   - curl --retry 10 --retry-delay 5 -v http://localhost:9200/
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Docker ELK stack
 
-[![Join the chat at https://gitter.im/deviantony/docker-elk](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/deviantony/docker-elk?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge) [![Elastic Stack version](https://img.shields.io/badge/ELK-5.6.1-blue.svg?style=flat)](https://github.com/deviantony/docker-elk/issues/173)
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -0,0 +1,37 @@
+sudo: required
+
+language: ruby
+
+services:
```
antoineco: Better avoid this, since you install Docker manually in the next steps.

.travis.yml
```diff
@@ -0,0 +1,37 @@
...
+before_script:
+  # Check docker & docker-compose versions
+  - docker --version
+  - docker-compose --version
+
+script:
+  - docker-compose build
+  - docker-compose up -d
+
+after_success:
```
antoineco: This should be moved to `script`. `after_success` does not fail on error.
https://docs.travis-ci.com/user/customizing-the-build/#Breaking-the-Build

README.md
```diff
@@ -1,6 +1,7 @@
 # Docker ELK stack
 
 [![Join the chat at https://gitter.im/deviantony/docker-elk](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/deviantony/docker-elk?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge) [![Elastic Stack version](https://img.shields.io/badge/ELK-5.6.2-blue.svg?style=flat)](https://github.com/deviantony/docker-elk/issues/175)
+[![Build Status](https://api.travis-ci.org/ivanrome/docker-elk.svg?branch=master)](https://travis-ci.org/ivanrome/docker-elk)
```
antoineco: Consider changing `ivanrome` to `deviantony` in the URL before we merge :)

antoineco:
@deviantony 

> Just a quick comment, what'd be the purpose of a travis automated buid here?

Travis CI would start the start and test that everything comes up as expected on every commit. Very handy to make sure an update didn't break I think.

deviantony:
Great, that'd be handy indeed !

.travis.yml
ivanrome: Language ruby is the dafault, avoiding this caused a validation warning in  https://lint.travis-ci.org

README.md
ivanrome: OK :smile:

.travis.yml
deviantony: What about using `language: generic` ?

See:

* https://github.com/travis-ci/travis-ci/issues/4895
* https://github.com/travis-ci/travis-ci/issues/2867#issuecomment-59262795

.travis.yml
ivanrome: OK :smile:  09d98784a5749fc488ec33a93ba2482c5b80a9f2

.travis.yml
ivanrome: OK 392b263c446f22b75216c21f908c29cd4c33fc1c

antoineco: I was more referring to the Docker service, which feels unnecessary here since you install it manually.

.travis.yml
```diff
@@ -0,0 +1,37 @@
+sudo: required
+
+language: generic
+
+services:
+  - docker
```
antoineco: As mentioned before, I believe docker is not necessary since you install it manually.

ivanrome: OK seems to work 4734c8e2d6413a1f776a466bae78f1e34df6d340

.travis.yml
```diff
@@ -0,0 +1,34 @@
...
+  # Check docker & docker-compose versions
+  - docker --version
+  - docker-compose --version
+
+script:
+  - docker-compose build
+  - docker-compose up -d
+
+  # Verifications
+  - sleep 4m
```
deviantony: 4 minutes to wait for the stack to be ready? I reckon `1m` should be enough.

.travis.yml
```diff
@@ -0,0 +1,34 @@
...
+  - docker network ls
```
deviantony: What's the point of the `docker network ls` command here?

.travis.yml
ivanrome: In the past I've tryed with lower values than ``4m``, like in this case (``3m``) :
https://travis-ci.org/ivanrome/docker-elk/builds/286973782

Note the failure of the build due to the error connecting with kibana
> The command "curl --retry 10 --retry-delay 5 -v http://localhost:5601/" exited with 52.
Note that in the past ``3m`` was enough time to start the container but not always. So I increased the value to ``4``

.travis.yml
ivanrome: Just to see some extra information after starting the containers.

Actually 4 containers are shown.

.travis.yml
deviantony: Ok, I forgot that it needs to build all the images first. Maybe upgrade that to `5m` then.

antoineco: Won't be necessary because the images are already built when `sleep` is executed.

.travis.yml
antoineco: Didn't you mean `docker-compose ps` if you want to see the containers running?

.travis.yml
ivanrome: Yeah, exactly, images are built before starting the container. What it takes time is starting the stack

.travis.yml
ivanrome: ``docker-compose  ps`` can be also added to show more informations of running containers.
``docker network`` allows to show other kind of information

antoineco: It shows
```$ docker network ls
NETWORK ID          NAME                DRIVER              SCOPE
d7d3e6672af7        bridge              bridge              local
fa39c8877bbd        dockerelk_elk       bridge              local
3fd9b4092147        host                host                local
0617e0553039        none                null                local
```
Is that really helpful?

deviantony: I don't think so. No point to show running container either, the healthcheck is the only interesting thing here.

antoineco: I would get the point of showing containers in some situations. Occasionally one of them terminates during startup, because some configuration is deprecated or multiple other reasons. Displaying the state of the stack would at least explain why the healthcheck fails, and help us pinpoint the failing service. What do you think @deviantony ?

deviantony: What about integrating `docker-compose logs` instead?

antoineco: You're right, makes more sense.

deviantony: I think that `docker-compose logs` would be enough to diagnose in case of issue.

ivanrome: Ok, if I undertand, I'll replace ``docker network`` by ``docker-compose logs``

.travis.yml
ivanrome: OK 2f58ba5f6157dcf21df49430eaad8e9850dca60d

antoineco:
Thanks a lot @ivanrome! Cool feature 👍

antoineco:
@deviantony you may have to enable the repo on https://travis-ci.org/deviantony/docker-elk :)

deviantony:
Thanks for this @ivanrome :+1: 

@antoineco I just did.

antoineco:
Beautiful :)

travis

ivanrome:
you are welcome

PR #187, opened 2017-10-22
Specify maintainer in Dockerfile

ivanrome:
As seen in https://www.codacy.com/app/ivanrome/docker-elk/issues

### DIFF
--- a/elasticsearch/Dockerfile
+++ b/elasticsearch/Dockerfile
@@ -1,5 +1,7 @@
 # https://github.com/elastic/elasticsearch-docker
 FROM docker.elastic.co/elasticsearch/elasticsearch:5.6.3
 
+LABEL maintainer="deviantony"
+
 # Add your elasticsearch plugins setup here
 # Example: RUN elasticsearch-plugin install analysis-icu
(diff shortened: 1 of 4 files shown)


antoineco:
Since the image is built by the end user and not the repo authors, shouldn't be/she the maintainer?

ivanrome:
It's true dockerfile is minimal and just a FROM clause exists.

Maybe the mainteiner should be elastic.co

antoineco:
Yes, it is already:
https://github.com/elastic/elasticsearch-docker/blob/master/templates/Dockerfile.j2#L109

ivanrome:
Yes, but that file is not in the project but in the source image.

deviantony:
I agree with @antoineco here, we provide simple wrapper around the official images in order to make the stack quite extensible. We don't really maintain any Docker images in here (nothing is hosted on any registry).

And if we wanted to add maintainers, I'd see my name as well as @antoineco's name (thanks again for the great job :+1:)

antoineco:
Closing this, but thanks for the proposal @ivanrome, hopefully it will come in handy for someone even if we don't merge it here.

PR #193, opened 2017-10-30
Fix link to default logstash config

lbausch:


logstash/config/logstash.yml
```diff
@@ -9,4 +9,4 @@ path.config: /usr/share/logstash/pipeline
 ## see https://www.elastic.co/guide/en/x-pack/current/xpack-settings.html
 ##     https://www.elastic.co/guide/en/x-pack/current/installing-xpack.html#xpack-enabling
 #
-xpack.monitoring.enabled: false
```
antoineco: Would it be possible to avoid this extra diff line?

lbausch: Sure!

antoineco:
Thanks @lbausch :+1:

PR #196, opened 2017-11-15
Updated the project for supporting the latest GA version 6.0.0 OSS Edition

OpenPj:
I have updated the ELK stack using the latest OSS edition but probably it could be useful to have different branches dedicated to the OSS and Enterprise editions.

Starting from this release to disable X-Pack we need to install the plugin and the disable it using properties. For Kibana and Logstash we can remove all the properties.

Hope this helps 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Docker ELK stack
 
 [![Join the chat at https://gitter.im/deviantony/docker-elk](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/deviantony/docker-elk?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
-[![Elastic Stack version](https://img.shields.io/badge/ELK-6.0.0-blue.svg?style=flat)](https://github.com/deviantony/docker-elk/issues/182)
+[![Elastic Stack version](https://img.shields.io/badge/ELK-6.0.0-blue.svg?style=flat)](https://github.com/deviantony/docker-elk/issues/196)
 [![Build Status](https://api.travis-ci.org/deviantony/docker-elk.svg?branch=master)](https://travis-ci.org/deviantony/docker-elk)
 
 Run the latest version of the ELK (Elasticsearch, Logstash, Kibana) stack with Docker and Docker Compose.
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ You can also choose to run it in background (detached mode):
 $ docker-compose up -d
 ```
 
-Give Kibana about 2 minutes to initialize, then access the Kibana web UI by hitting
+Give Kibana a few seconds to initialize, then access the Kibana web UI by hitting
 [http://localhost:5601](http://localhost:5601) with a web browser.
 
 By default, the stack exposes the following ports:
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Docker ELK stack
 
 [![Join the chat at https://gitter.im/deviantony/docker-elk](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/deviantony/docker-elk?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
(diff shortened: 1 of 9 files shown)


antoineco:
Thanks for the PR! The docs still mention that X-Pack options should be explicitly enabled/disabled: https://www.elastic.co/guide/en/x-pack/current/xpack-settings.html. Could you clarify why Kibana and Logstash do not need it anymore?

josephholsten:
Do we really want to use the OSS images without x-pack? I’m managing security externally, so it doesn’t bother me.

I do wonder if we document how to migrate kinsman indexes.

josephholsten:
...if we /should/ document how...

antoineco:
It makes sense to use OSS on the `master` branch if it doesn't include X-Pack (question mark here), especially since we have a dedicated branch for X-Pack.

OpenPj:
> Thanks for the PR! The docs still mention that X-Pack options should be explicitly enabled/disabled: https://www.elastic.co/guide/en/x-pack/current/xpack-settings.html. Could you clarify why Kibana and Logstash do not need it anymore?

It seems that only ElasticSearch requires to install the X-Pack plugin and it requires also to mention all the properties for disabling it. 

For Kibana and Logstash there is no need to install and disable the plugin. You can simply avoid to install the plugin and you don't need to add properties in the configuration files for disabling X-Pack.

antoineco:
@OpenPj it seems like Elasticsearch also has an `oss` flavour 🏆 
https://github.com/elastic/elasticsearch-docker/blob/master/Makefile#L16

README.md
```diff
@@ -1,7 +1,7 @@
 # Docker ELK stack
 
 [![Join the chat at https://gitter.im/deviantony/docker-elk](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/deviantony/docker-elk?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
-[![Elastic Stack version](https://img.shields.io/badge/ELK-5.6.3-blue.svg?style=flat)](https://github.com/deviantony/docker-elk/issues/182)
+[![Elastic Stack version](https://img.shields.io/badge/ELK-6.0.0-blue.svg?style=flat)](https://github.com/deviantony/docker-elk/issues/182)
```
antoineco: Line 129 also contains a reference to the current version.

elasticsearch/Dockerfile
```diff
@@ -1,5 +1,6 @@
 # https://github.com/elastic/elasticsearch-docker
-FROM docker.elastic.co/elasticsearch/elasticsearch:5.6.3
+FROM docker.elastic.co/elasticsearch/elasticsearch-oss:6.0.0
 
 # Add your elasticsearch plugins setup here
 # Example: RUN elasticsearch-plugin install analysis-icu
+RUN elasticsearch-plugin install x-pack
```
antoineco: Why do you install X-pack just to disable it the config?

elasticsearch/config/elasticsearch.yml
```diff
@@ -19,7 +19,8 @@ discovery.type: single-node
 ## see https://www.elastic.co/guide/en/x-pack/current/xpack-settings.html
 ##     https://www.elastic.co/guide/en/x-pack/current/installing-xpack.html#xpack-enabling
 #
-xpack.security.enabled: false
```
antoineco: See previous comment, we should get rid of all `xpack.*` settings.

elasticsearch/Dockerfile
OpenPj: Because otherwise ElasticSearch returns an exception during the bootstrap and the application can't be started. I sincerely don't know why and it sounds to me just a little bit strange.

antoineco: That's because you didn't delete `xpack.*` settings, so ES crashes due to unsupported options.

antoineco:
@OpenPj I pushed a couple of changes and merged `master` in your fork. Good to go!

PR #204, opened 2017-11-30
Make config volumes read-only

bean5:
Because the fs doesn't need to be write-able.


antoineco:
Makes sense, thanks @bean5 :)

Issue #205, opened 2017-12-01
Broken compatibility with Docker < 17.03 


atais:
Hello

I start with `docker-compose up` and this is what I get

```
[/location] # docker-compose up
Building elasticsearch
Step 1 : ARG ELK_VERSION
ERROR: Service 'elasticsearch' failed to build: Please provide a source image with `from` prior to commit
```
My system
```
[/location] # docker --version
Docker version 1.11.2, build 781516c
[/location] # docker-compose --version
docker-compose version 1.13.0, build 1719ceb
```

antoineco:
@atais this is a known issue with older Docker versions, apologies for that :(

Docker 17.03 is currently required because the `FROM` statements we use inside our Dockerfiles contain a variable.

As a workaround can you try to replace that variable with the actual version number (6.0.0) in all 3 Dockerfiles?

I'm already considering to revert the commit that introduced this change, you're not the first person to report it.

atais:
I would totally use newest CE version of docker but I am using my QNAP (File Server) with build-in docker and I can not update the version as I like 😢 
Thanks for the solution!

deviantony:
> I'm already considering to revert the commit that introduced this change, you're not the first person to report it.

Agree, we should still be able to support older Docker versions (or at least update the requirements in the README).

atais:
Created a pull request

PR #206, opened 2017-12-01
fixing broken compatibility with Docker < 17.03

atais:
Looking at lines statistics I don't think it was worth adding 😄 

Tested on 
```
[/location] # docker --version
Docker version 1.11.2, build 781516c
[/location] # docker-compose --version
docker-compose version 1.13.0, build 1719ceb
```

Ref:  #205

### DIFF
--- a/.env
+++ b/.env
@@ -1,1 +1,0 @@
-ELK_VERSION=6.0.0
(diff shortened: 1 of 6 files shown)


antoineco:
Thanks! I amended your commit and removed the changes to the README as well.

PR #202, opened 2017-11-24
Full upgrade for Vagrant

dulin:
I merged everything from master, updated Ubuntu to 16.04, updated docker
fixed issue with:
> [1]: max virtual memory areas vm.max_map_count [65530] is too low, increase to at least [262144]

Closes #199

PR #217, opened 2017-12-25
Correct used ELK versions

jemacom:
Both vagrant and searchguard branches still using ELK version 5


antoineco:
Thanks, that's a valid point.

Searchguard is now up-to-date and the Vagrant setup is also being dealt with in #202.

antoineco:
The `vagrant` branch is now updated as well 👍

PR #240, opened 2018-02-05
Mount searchguard config directory

dgptamayo:


antoineco:
Wrong branch

docker-compose.yml
```diff
@@ -7,6 +7,7 @@ services:
       context: elasticsearch/
     volumes:
       - ./elasticsearch/config/elasticsearch.yml:/usr/share/elasticsearch/config/elasticsearch.yml:ro
+      - ./elasticsearch/config/sg:/usr/share/elasticsearch/config/sg:ro
```
antoineco: This is not enough, you should also remove all related statements from the Dockerfile.

PR #246, opened 2018-02-14
making docker-compose version 3

gekitsuu:
I'm using this as a stack in docker swarm so I made this work for docker-compose version 3


gekitsuu:
I will make this a separate file named docker-compose-swarm.yml so we don't have to run docker in swarm mode on travis

antoineco:
Since I don't see anything Swarm specific here, shall we just convert the current Compose file?

gekitsuu:
Fixed it up to be version 3 and work with travis.

docker-compose.yml
```diff
@@ -43,6 +46,5 @@ services:
       - elasticsearch
 
 networks:
-
   elk:
-    driver: bridge
+    driver: overlay
```
deviantony: Forcing the driver type to `overlay` will make this stack compatible with Swarm only.

docker-compose.yml
```diff
@@ -1,8 +1,9 @@
-version: '2'
+version: '3'
 
 services:
 
   elasticsearch:
+    image: dockerelk_elasticsearch:latest
     build:
```
deviantony: the `build` option is ignored when using the `docker stack` command.

docker-compose.yml
```diff
@@ -16,6 +17,7 @@ services:
       - elk
 
   logstash:
+    image: dockerelk_logstash:latest
```
deviantony: I don't like the fact that we need to rely on `docker-compose` first to build the images. `docker-compose` and `docker stack` are two different tools with different purposes and there should not be any dependency on one or the other.

gekitsuu:
I'm happy to split it into another file and it seems to work 👌 with the repository images.

docker-swarm.yml
```diff
@@ -0,0 +1,43 @@
...
+    volumes:
+      - ./logstash/config/logstash.yml:/usr/share/logstash/config/logstash.yml:ro
+      - ./logstash/pipeline:/usr/share/logstash/pipeline:ro
+    ports:
+      - "5000:5000"
+    environment:
+      LS_JAVA_OPTS: "-Xmx256m -Xms256m"
+    networks:
+      - elk
+    depends_on:
```
deviantony: The `depends_on` option is ignored when using the `docker stack` command.

docker-swarm.yml
```diff
@@ -0,0 +1,43 @@
...
+
+  kibana:
+    image: docker.elastic.co/kibana/kibana-oss:6.2.1
+    volumes:
+      - ./kibana/config/:/usr/share/kibana/config:ro
+    ports:
+      - "5601:5601"
+    networks:
+      - elk
+    depends_on:
```
deviantony: The `depends_on` option is ignored when using the `docker stack` command.

docker-swarm.yml
```diff
@@ -0,0 +1,43 @@
+version: '3'
+
+services:
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch-oss:6.2.1
+    volumes:
+      - ./elasticsearch/config/elasticsearch.yml:/usr/share/elasticsearch/config/elasticsearch.yml:ro
```
deviantony: I believe that this will work on a single node cluster but will probably fail on a cluster with nodes > 1.

A solution here would be to define the equivalent of the configuration file as environment variables.

docker-swarm.yml
```diff
@@ -0,0 +1,43 @@
...
+      - "9300:9300"
+    environment:
+      ES_JAVA_OPTS: "-Xmx256m -Xms256m"
+    networks:
+      - elk
+
+  logstash:
+    image: docker.elastic.co/logstash/logstash-oss:6.2.1
+    volumes:
+      - ./logstash/config/logstash.yml:/usr/share/logstash/config/logstash.yml:ro
```
deviantony: I believe that this will work on a single node cluster but will probably fail on a cluster with nodes > 1.

A solution here would be to define the equivalent of the configuration file as environment variables.

docker-swarm.yml
```diff
@@ -0,0 +1,43 @@
...
+      LS_JAVA_OPTS: "-Xmx256m -Xms256m"
+    networks:
+      - elk
+    depends_on:
+      - elasticsearch
+
+  kibana:
+    image: docker.elastic.co/kibana/kibana-oss:6.2.1
+    volumes:
+      - ./kibana/config/:/usr/share/kibana/config:ro
```
deviantony: I believe that this will work on a single node cluster but will probably fail on a cluster with nodes > 1.

A solution here would be to define the equivalent of the configuration file as environment variables.

.travis.yml
```diff
@@ -32,3 +32,11 @@ script:
   - docker-compose logs
   - curl --retry 10 --retry-delay 5 -D- http://localhost:9200/
   - curl --retry 10 --retry-delay 5 -D- http://localhost:5601/api/status
+  - docker-compose down && docker-compose rm
+
+  # Test docker-swarm.yml
+  - docker swarm init
```
deviantony: Cannot confirm that this will work, @antoineco thoughts?

antoineco: It should, at least the syntax is correct. I know nothing about docker stacks though. Isn't the file format supposed to be different from Compose?

deviantony: All good then.

Actually it's the same file format but not the same version (`docker-compose` uses `2.x` and `docker stack` uses `3.x`).
See the documentation here: https://docs.docker.com/compose/compose-file/

antoineco: I got mixed up with Cloud stacks then
https://docs.docker.com/docker-cloud/apps/stack-yaml-reference/

.travis.yml
```diff
@@ -32,3 +32,11 @@ script:
...
+  - docker stack deploy -c ./docker-swarm.yml elk
+  - sleep 30
```
antoineco: Is there any way to display the stack logs?

docker-swarm.yml
antoineco: Agree, our current approach of mounting files was meant to make it easy to fiddle with configurations, locally. It doesn't apply very well to a distributed infrastructure.

For the sake of consistency it would be nice to move everything to env vars, unfortunately only a subset of variables is supported.

deviantony: Yup, I wonder if the defaults configuration variables we ship in configuration are part of that supported subset.

antoineco: ES definitely, Kibana and Logstash possibly. Search Guard unlikely :)

deviantony: I was thinking about adding this setup in the master branch only to start with. We'll think about other setups after that :-)

.travis.yml
deviantony: Not using a single command, we can display the logs of each services in the stack using the `docker service logs` command though

.travis.yml
```diff
@@ -32,3 +32,11 @@ script:
   - docker-compose logs
   - curl --retry 10 --retry-delay 5 -D- http://localhost:9200/
   - curl --retry 10 --retry-delay 5 -D- http://localhost:5601/api/status
+  - docker-compose down && docker-compose rm
```
deviantony: I believe `docker-compose down` is enough, no need for `&& docker-compose rm`

gekitsuu:
Sorry for going quiet about this, I'm moving across the country in a couple weeks so might be awhile before I can address the issues.

hastarin:
You could use 3.3 and the configs option for handling the configuration files:
https://docs.docker.com/compose/compose-file/#configs

deviantony:
@hastarin that might be the way to go for configuration files, yes !

hastarin:
I've done a 3.3 version here that can hopefully help with this:
https://github.com/hastarin/docker-elk/blob/master/docker-stack.yml

I'm using a variation of that in a test environment.

deviantony:
@hastarin looks great ! Mind opening a PR to replace this one?

.travis.yml
antoineco: Even better, `down -v` to also wipe volumes 🗑

deviantony: Agree.

Note, I'd favor the implementation in https://github.com/deviantony/docker-elk/pull/252 over this PR. Will ask for the changes in the `.travis.yml` file in #252

PR #252, opened 2018-03-09
Add a docker-stack.yml to support a basic Docker Swarm deployment

hastarin:
NOTE: I only have this running in a test environment and haven't done much with it yet but creating a pull request since @deviantony asked.

Closes #246

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ Based on the official Docker images:
 1. [Requirements](#requirements)
    * [Host setup](#host-setup)
    * [SELinux](#selinux)
+   * [DockerForWindows](#dockerforwindows)
 2. [Getting started](#getting-started)
    * [Bringing up the stack](#bringing-up-the-stack)
    * [Initial setup](#initial-setup)
@@ -60,6 +61,10 @@ apply the proper context:
 ```console
 $ chcon -R system_u:object_r:admin_home_t:s0 docker-elk/
 ```
+
+### DockerForWindows
+
+If you're using Docker for Windows, ensure the 'Shared Drives' feature is enabled for the C: drive (Docker for Windows > Settings > Shared Drives). [MSDN article detailing Shared Drives config](https://blogs.msdn.microsoft.com/stevelasker/2016/06/14/configuring-docker-for-windows-volumes/).
 
 ## Usage
 


deviantony:
Thanks, I'll try to have a look at this shortly.

elasticsearch/config/elasticsearch.yml
```diff
@@ -13,4 +13,4 @@ discovery.zen.minimum_master_nodes: 1
 ## Use single node discovery in order to disable production mode and avoid bootstrap checks
 ## see https://www.elastic.co/guide/en/elasticsearch/reference/current/bootstrap-checks.html
 #
-discovery.type: single-node
+discovery.type: zen
```
deviantony: @antoineco Do you think that might introduce issues with the standard deployment method (e.g. compose)?

hastarin: Sorry that's a leftover from my local experiments.  I'll remove that.

docker-stack.yml
```diff
@@ -0,0 +1,74 @@
+version: '3.3'
+
+networks:
+  net:
```
deviantony: For consistency with the compose file, please rename the network to `elk`.

docker-stack.yml
```diff
@@ -0,0 +1,74 @@
...
+      - source: logstash_pipeline
+        target: /usr/share/logstash/pipeline/logstash.conf
+    environment:
+      LS_JAVA_OPTS: "-Xmx256m -Xms256m"
+    networks:
+      - net
+    deploy:
+      mode: replicated
+      replicas: 1
+      resources:
```
deviantony: I don't know if the resources limit are actually required here. This project aims to provide a generic solution, we might want to let the user actually define these.

docker-stack.yml
```diff
@@ -0,0 +1,74 @@
...
+      - "5601:5601"
+    configs:
+      - source: kibana_config
+        target: /usr/share/kibana/config/kibana.yml
+    networks:
+      - net
+    deploy:
+      mode: replicated
+      replicas: 1
+      resources:
```
deviantony: I don't know if the resources limit are actually required here. This project aims to provide a generic solution, we might want to let the user actually define these.

docker-stack.yml
hastarin: Are you sure?  If you deploy the stack as "elk" you will then end up with a network named "elk_elk".

docker-stack.yml
hastarin: You have JAVA limitations passed in via environment variables.  Wouldn't it make sense to provide something similar for the docker container itself?   Users can easily edit or remove these but if you're new to docker as I am you may not even know they exist without seeing them there.

hastarin:
I know next to nothing about Travis CI.   NOTE:  Even locally using "localhost" does not work for me.  Using the hostname or 127.0.0.1 does.  If this latest build doesn't pass I don't know what more I can try without spending more time than I have available on it.

docker-stack.yml
deviantony: Sorry for the late update @hastarin 

Yes, it's all right, please rename it to `elk`.

deviantony:
@antoineco any hint on the Travis issue?

antoineco:
@deviantony looks like a benign timeout issue to me, Swarm may take slighly longer than Compose to start services, let me check.

antoineco:
That was it. I increased the sleep to 40sec and the test plan succeeds 👍

PR #257, opened 2018-03-22
Define persistent elk network name

jhoelzl:
Otherwise network name is composed of the current folder name + "_elk". This could differ through various installations.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '2'
+version: '3.5'
 
 services:
 
@@ -46,3 +46,4 @@ networks:
 
   elk:
     driver: bridge
+    name: elk_network


antoineco:
Travis failed with `Version in "./docker-compose.yml" is unsupported.`. See my comment.

jhoelzl:
Unfortunately, the version 3.5 is necessary to define the network name: https://docs.docker.com/compose/compose-file/#name-1

antoineco:
From what I see this requires Compose `1.19` and Docker `17.06`. Last time we introduced a change which required such recent components we broke multiple users' setup, according to the feedback we received.

Not sure what version we commit to supporting. The latest 3, 4, 5 major versions? @deviantony is the decision taker here :)

deviantony:
> Otherwise network name is composed of the current folder name + "_elk". This could differ through various installations.

We do not relay on the network name so it does not really matter that the network name actually differs for different deployments.

I agree that this change makes deployment cleaner but I don't think it's worth removing support for older compose/Docker version just for this enhancement.

Issue #229, opened 2018-01-15
Using with Docker stack in Swarm mode


yogeek:
Hello,

Thank you for sharing this. Is it possible to use the same mechanism but to scale it in a Swarm mode cluster using the 'global' deploy mode with the `docker stack deploy` command instead of docker-compose on a single host ? I would allow each new node to have a new ES service added to the cluster.
I tried to deploy it but it seems that putting the service name as `discovery.zen.ping.unicast.hosts: elasticsearch` is not sufficient for ES to detect the other nodes in swarm mode. Here is the error log I have : 
```
java.net.UnknownHostException: elasticsearch: Name or service not known
```

And here is the docker-stack.yml I use : 
```
version: "3"

services:

  elasticsearch:
    image: docker.elastic.co/elasticsearch/elasticsearch:5.6.5
    ports:
      - "9200:9200"
      - "9300:9300"
    environment:
      transport.host: 0.0.0.0
      discovery.zen.ping.unicast.hosts: elasticsearch
      discovery.zen.minimum_master_nodes: 1
      cluster.name: elasticsearch
      ES_JAVA_OPTS: "-Xms512m -Xmx512m -XX:-AssumeMP"
      xpack.security.enabled: "false"
      xpack.monitoring.enabled: "false"
      xpack.ml.enabled: "false"
      xpack.graph.enabled: "false"
      xpack.watcher.enabled: "false"
      http.cors.enabled: "true"
      http.cors.allow-origin: "*"
    ulimits:
      memlock:
        soft: -1
        hard: -1
      nofile:
        soft: 65536
        hard: 65536
    volumes:
      - elasticsearch_data:/usr/share/elasticsearch/data
    networks:
      spirit1-apis:
    deploy:
      resources:
        limits:
          memory: 1G
      mode: global
      update_config:
        parallelism: 1
        delay: 1m30s
      restart_policy:
        condition: on-failure

volumes:
  elasticsearch_data:
```

Why can't the DNS round robin of Swarm cannot apply here as it applies in your local docker-compose stack  ?

Thank you

antoineco:
From your description it's unclear how the `spirit1-apis` network is configured, is it an overlay network?

Also you'll probably want to set `endpoint mode: dnsrr` in your `deploy` block, otherwise you'll only get a single IP for all ES nodes so the unicast discovery won't work.

antoineco:
It also seems like you have to prepend your service name with `task.` in order to get a response from the DNS.

[...]

PR #262, opened 2018-04-03
Add a docker-stack.yml to support a basic Docker Swarm deployment

antoineco:
Redo of #252 because I pushed to the wrong remote and broke the branch 🙈 
Closes #246


.travis.yml
```diff
@@ -5,30 +5,53 @@ language: generic
...
-script:
+before_script:
+  # Build images
   - docker-compose build
-  - docker-compose up -d
 
-  # Verifications
-  - sleep 30
+  # Pre-pull buildpack image (curl 7.52.0+)
+  - docker pull buildpack-deps:artful-curl
```
deviantony: What is the purpose of this image?

docker-stack.yml
```diff
@@ -0,0 +1,83 @@
+version: '3.3'
+
+networks:
+  net:
```
deviantony: Just wanted to rename the network to `elk` to stay consistent with our compose deployment.

docker-stack.yml
```diff
@@ -0,0 +1,83 @@
+version: '3.3'
```
deviantony: Not sure that we're using any 3.3 specific features here, might want to use 3 instead.

.travis.yml
antoineco: It contains a version of `curl` that supports `--retry-connrefused`, which is unfortunately not available on Ubuntu Trusty (base OS of Travis workers).

docker-stack.yml
antoineco: Seems like it is, Travis just failed with
```
$ docker stack deploy -c ./docker-stack.yml elk
configs Additional property configs is not allowed
```
Reverting.

antoineco: Found this in the documentation @ https://docs.docker.com/compose/compose-file/#short-syntax:
> Note: `config` definitions are only supported in version 3.3 and higher of the compose file format.

deviantony: Yup, we should stay on 3.3.

docker-stack.yml
```diff
@@ -0,0 +1,83 @@
...
+      - source: logstash_pipeline
+        target: /usr/share/logstash/pipeline/logstash.conf
+    environment:
+      LS_JAVA_OPTS: "-Xmx256m -Xms256m"
+    networks:
+      - elk
+    deploy:
+      mode: replicated
+      replicas: 1
+# Uncomment the following to limit the memory used by the container
```
deviantony: I believe that we should remove the comments regarding resource limitations.

deviantony:
Gave it a quick try, was not able to inject data into the stack, will retry and investigate later.

antoineco:
If I remember well (https://github.com/deviantony/docker-elk/issues/229#issuecomment-360573244) services deployed in a Swarm do not resolve to their service name (eg. `elasticsearch`). Instead, the service name must be prepended with `tasks.` (eg. `tasks.elasticsearch`).

antoineco:
I added a test which injects dummy data into Logstash and checks whether the document ended up in Elasticsearch.

DavidJBianco:
I tried this in a three-node swarm, and ```docker-swarm.yml``` doesn't seem to work as-is.  The swarm starts, but if you list nodes, you'll only see a single node, even though multiple containers run.  For some reason I don't understand, they all publish the same IP, so the cluster thinks it has only one node. 

I modified the ```elasticsearch``` stanza to the following  and so far it's working well:

```
  elasticsearch:
    image: docker.elastic.co/elasticsearch/elasticsearch-oss:6.2.2
    ports:
      - "9200:9200"
      - "9300:9300"
    configs:
      - source: elastic_config
        target: /usr/share/elasticsearch/config/elasticsearch.yml
    environment:
      ES_JAVA_OPTS: "-Xmx256m -Xms256m"
      network.bind_host: 0.0.0.0
      network.publish_host: _eth0_
      discovery.type: zen
      discovery.zen.ping.unicast.hosts: tasks.elasticsearch
    networks:
      - elk
    deploy:
      mode: global
```

You may need to change ```_eth0_``` depending on your NIC.  

Maybe you prefer not to use "global" deployment mode, but it makes it much easier to scale the cluster.

antoineco:
@DavidJBianco 
> but if you list nodes, you'll only see a single node

You're talking about Elasticsearch, right?

antoineco:
That's a good point. However I believe manually increasing the value of the `vm.max_map_count` kernel parameter is a prerequisite to using `zen` discovery isn't it? (cf. https://github.com/deviantony/docker-elk/wiki/Elasticsearch-cluster)

DavidJBianco:
Yeah, it is.  I already had that set on my Docker hosts, so I didn't include that, but you're right.

antoineco:
@deviantony I finally updated the README and rebased on `master`.
I also tried scaling ES and can confirm it works:

```console
$ docker stack services elk   
ID                  NAME                MODE                REPLICAS            IMAGE                                                     PORTS
yh97pxidx7y0        elk_elasticsearch   replicated          2/2                 docker.elastic.co/elasticsearch/elasticsearch-oss:6.4.0   *:9200->9200/tcp, *:9300->9300/tcp
ewaq0ejdi9ct        elk_kibana          replicated          1/1                 docker.elastic.co/kibana/kibana-oss:6.4.0                 *:5601->5601/tcp
ykd955pfqliy        elk_logstash        replicated          1/1                 docker.elastic.co/logstash/logstash-oss:6.4.0             *:5000->5000/tcp, *:9600->9600/tcp
```

```console
$ curl localhost:9200/_nodes              
{"_nodes":{"total":2,"successful":2,"failed":0}, ...
```

OK to merge?

deviantony:
@antoineco awesome :-)

PR #276, opened 2018-05-22
Elk with Searchguard and Sentinl

parivero:
Hi,
I've extended your docker-elk repo by adding support for Sentinl. Be my guest to add this demo to your repo if you find it useful.

Regards.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Run the latest version of the ELK (Elasticsearch, Logstash, Kibana) stack with D
 
 **NOTE:** This version has:
 * [Search Guard support](https://github.com/floragunncom/search-guard).
-* [Sentinel](https://github.com/sirensolutions/sentinl).
+* [Sentinl](https://github.com/sirensolutions/sentinl).
 
 It will give you the ability to analyze any data set by using the searching/aggregation capabilities of Elasticsearch
 and the visualization power of Kibana.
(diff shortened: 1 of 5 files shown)


parivero:
Hi, I think it could be a new branch like SearchGuard and it would be a free approach to the ElasticSearch x-pack (which includes security, alerts and reports). In the future, someone could add the rest of the x-pack features.
In any case i can gladly create a new page in the wiki.

antoineco:
The problem with new branches is that they have to be kept in sync manually by maintainers, because they can't be merged into one another.
This project tries to remain quite bare so that people can fork it and add their own plugins - like you did - without changing the Compose file or restructuring the repo. It's a boilerplate to build upon.
As nice as it would be to have one dedicated branch per popular plugin, it's simply not realistic without letting some branches rot over time (cf the `vagrant` branch).

If you add a Markdown file to this PR reusing the instructions and examples you already wrote, I will gladly add it to the Wiki for you.

parivero:
First sorry for the delay. The file sentinl.md was added to PR.

Regards.

antoineco:
@parivero It took me a while but I finally created the page and added your content to it (slightly rephrased): https://github.com/deviantony/docker-elk/wiki/Popular-integrations

Thanks a lot! 🙌 
Hopefully we'll see some other similar contributions :)

PR #297, opened 2018-07-21
Add curator

tecris:
This pull request adds curator (running in a docker container as all other parts of the ELK stack).
The image allows for following environment variables:
- elasticsearch host (ELASTICSEARCH_HOST)
- cron configuration (CRON)
- curator config file (CONFIG_FILE)
- curator action file (COMMAND)
- curator filtertype age [unit count](https://www.elastic.co/guide/en/elasticsearch/client/curator/current/fe_unit_count.html) (UNIT_COUNT)


curator/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM alpine:latest
```
antoineco: Please pin the Alpine version to avoid future package incompatibilities.

curator/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN apk --update add python py-pip && \
+    pip install elasticsearch-curator && \
+     rm -rf /var/cache/apk/*
+
+ADD entrypoint.sh /entrypoint.sh
+
+WORKDIR /usr/share/curator
+COPY config ./config
+
+RUN chmod +x /entrypoint.sh
```
antoineco: I'd rather set +x on the source file, this way you could remove that line.

curator/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM alpine:latest
+
+
+RUN apk --update add python py-pip && \
+    pip install elasticsearch-curator && \
```
antoineco: Formatting a bit off, consider using the following format used in official Dockerfiles:
```dockerfile
RUN foo \
<tab> && bar
```

curator/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+     rm -rf /var/cache/apk/*
+
+ADD entrypoint.sh /entrypoint.sh
```
antoineco: ADD should be reserved for copying files from HTTP URLs. Besides you could simply copy to `/` since you don't rename the file.

curator/entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+echo "$CRON /usr/bin/curator  --config ${CONFIG_FILE} ${COMMAND}" >>/etc/crontabs/root
+
+crond -f -d 8 -l 8
```
antoineco: `crond` doesn't handle SIGTERM properly. Consider monitoring it using `tini`.

curator/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM alpine:latest
+
+
+RUN apk --update add python py-pip && \
+    pip install elasticsearch-curator && \
+     rm -rf /var/cache/apk/*
```
antoineco: Please use `apk --no-cache` so you can avoid that line.

docker-compose.yml
```diff
@@ -42,6 +42,20 @@ services:
     depends_on:
       - elasticsearch
 
+  curator:
```
antoineco: Please move this service to https://github.com/deviantony/docker-elk/blob/master/extensions/.

antoineco:
By the way, why not using the [Dockerfile provided by the curator project](https://github.com/elastic/curator/blob/master/Dockerfile)?

extensions/curator/config/delete_log_files_curator.yml
```diff
@@ -4,10 +4,10 @@ actions:
     description: >-
       Delete indices.  Find which to delete by first limiting the list to logstash-
```
antoineco: Extraneous space after "indices."

extensions/curator/config/delete_log_files_curator.yml
```diff
@@ -4,10 +4,10 @@ actions:
...
       prefixed indices. Then further filter those to prevent deletion of anything
-      less than 3 days old. Next filter by space, to those indices in excess of
-      3g of usage.
+      less than ${UNIT_COUNT} days old.Ignore the error if the filter does not result in an
```
antoineco: Missing space after "old."

extensions/curator/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM alpine:3.8
 
 RUN apk --update add --no-cache python py-pip \
-  && pip install elasticsearch-curator
+  && pip install elasticsearch-curator==5.5.4
```
antoineco: Excellent idea. Would you mind setting the version using an environment variable to make it clearer? [example](https://github.com/deviantony/docker-elk/blob/searchguard/elasticsearch/Dockerfile#L8-L10)

extensions/curator/curator-docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+version: '2'
+
+services:
+
+
+  curator:
+    build:
+      context: extensions/curator/
+    environment:
+      ELASTICSEARCH_HOST: "elasticsearch"
```
antoineco: Best practice: avoid quotes in YAML when not necessary, double quotes are especially evil in some circumstances.

extensions/curator/curator-docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+      CRON: "0 0 * * *"
+      CONFIG_FILE: "/usr/share/curator/config/curator.yml"
```
antoineco: Would relative paths work since you changed the container's workdir in the Dockerfile?

extensions/curator/entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+echo "$CRON /usr/bin/curator  --config ${CONFIG_FILE} ${COMMAND}" >>/etc/crontabs/root
+
+crond -f -d 8 -l 8
```
antoineco: You defined `CRON: 0 0 * * *` in the config file, so do we need to run `crond` at all?

If yes: `crond` doesn't handle SIGTERM properly. Consider monitoring it using [`tini`](https://pkgs.alpinelinux.org/package/edge/community/x86/tini). eg.
```sh
tini -- crond ...
```

tecris: It really depends if curator needs to run automatically or manually.
As for tini, I need to do basic learning/reading.

extensions/curator/curator-docker-compose.yml
```diff
@@ -7,7 +7,7 @@ services:
     build:
       context: extensions/curator/
     environment:
-      ELASTICSEARCH_HOST: "elasticsearch"
+      ELASTICSEARCH_HOST: elasticsearch
       CRON: "0 0 * * *"
```
antoineco: Quotes

extensions/curator/curator-docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+
+services:
+
+
+  curator:
+    build:
+      context: extensions/curator/
+    environment:
+      ELASTICSEARCH_HOST: elasticsearch
+      CRON: "0 0 * * *"
```
antoineco: Quotes

extensions/curator/curator-docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+      CONFIG_FILE: "/usr/share/curator/config/curator.yml"
```
antoineco: Quotes

extensions/curator/curator-docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+      COMMAND: "/usr/share/curator/config/delete_log_files_curator.yml"
```
antoineco: Quotes

extensions/curator/entrypoint.sh
antoineco: tini is really dumb, it doesn't have any option. All you have to do is call it as I suggested and it will take care of starting `crond` as a child process and pass it all OS signals received by the container.

As a bonus, it will reap zombie processes that may end up staying around forever in your container :)

The best reference I could this about is https://github.com/krallin/tini/blob/master/README.md

You will also need to enable the community repo: https://wiki.alpinelinux.org/wiki/Enable_Community_Repository

```dockerfile
RUN echo 'http://dl-6.alpinelinux.org/alpine/edge/community' > /etc/apk/repositories \
  && apk -U add --no-cache tini python...
```

tecris:
Thanks for sharing your work :), happy to contribute.

PR #303, opened 2018-08-06
It is now possible to set the ELK version in the .env file.

ffknob:
I found it easier to change the version of the ELK stack this way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,12 +43,14 @@ Based on the official Docker images from Elastic:
 6. [JVM tuning](#jvm-tuning)
    * [How can I specify the amount of memory used by a service?](#how-can-i-specify-the-amount-of-memory-used-by-a-service)
    * [How can I enable a remote JMX connection to a service?](#how-can-i-enable-a-remote-jmx-connection-to-a-service)
+7. [Updates](#updates)
+   * [Using a newer stack version](#using-a-newer-stack-version)
 
 ## Requirements
 
 ### Host setup
 
-1. Install [Docker](https://www.docker.com/community-edition#/download) version **1.10.0+**
+1. Install [Docker](https://www.docker.com/community-edition#/download) version **17.03+**
 2. Install [Docker Compose](https://docs.docker.com/compose/install/) version **1.6.0+**
 3. Clone this repository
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -66,7 +66,11 @@ $ chcon -R system_u:object_r:admin_home_t:s0 docker-elk/
 
 If you're using Docker for Windows, ensure the "Shared Drives" feature is enabled for the `C:` drive (Docker for Windows > Settings > Shared Drives). See [Configuring Docker for Windows Shared Drives](https://blogs.msdn.microsoft.com/stevelasker/2016/06/14/configuring-docker-for-windows-volumes/) (MSDN Blog).
 
-## Usage
+## Getting started
+
+### Version
+
+Change the ELK stack version in the _.env_ file.
```
antoineco: I'd rather move that to the customization section. The easiest would be to revert this: https://github.com/deviantony/docker-elk/commit/8f5b242d6a51615899ae920cc420b6393853ebcd

antoineco:
Good idea! Fast forward 8 months, I believe we can now assume users should have a recent enough Docker version. @deviantony, concerns?

deviantony:
Sounds good to me, maybe rename `TAG` to something more explicit.

ffknob:
How about ELK_TAG instead of TAG?
Or do you think it makes sense to have differente TAG variable to the different Elastic product (ELASTICSEARCH_TAG, LOGSTASH_TAG, KIBANA_TAG, ...) so it will be possible to have different versions for different products...?

deviantony:
I believe that we should use the same version for each component of the stack, this is the way Elastic wants their stack to be used. I was just thinking to rename it to something else like `VERSION` as it's  more explicit IMO.

Thoughts @antoineco

antoineco:
I actually liked the original commit by the original author :)
https://github.com/deviantony/docker-elk/commit/8f5b242d6a51615899ae920cc420b6393853ebcd
Why not just reverting it?

deviantony:
@antoineco Agree

antoineco:
Commit reverted, merging.

PR #314, opened 2018-09-02
Update docker-compose.yml

chadyred:


chadyred:
For readability, I do this automatically. So, I do a better commit message.

antoineco:
I appreciate the suggestion but this sounds more like a matter of personal taste than a actual improvement to me. Sections were separated by new lines conscientiously and I don't find the current file particularly unreadable.

PR #315, opened 2018-09-11
Add gelf input to logstash

SimonSchneider:


antoineco:
@SimonSchneider thanks for your contribution, but the correct way to customize your stack is to fork this repo and implement your configuration in your own repository.

We can't accept configuration changes otherwise nobody would be able to agree on what's required or not (and we have a lot of users). Instead, we ship a minimal configuration that everyone can use as a base.

PR #324, opened 2018-09-18
fix docker image name used in docker-stack.yml

souhaiebtar:


antoineco:
Tests are failing, for this to work you will have to update the Compose version in `.travis.yml` as well.

souhaiebtar:
@antoineco after viewing your last push `Fix failing CI test`,  i found that the change i added were there before you removed them to fix travis build, i'm not familiar with travis, this is why i will close the comment until i found a solution

antoineco:
Yes but to be honest I haven't managed to fix the issue myself, the naming was one thing but there is another issue with Logstash apparently.

souhaiebtar:
i will try  to build using the same build envrionment as travis (using vagrant), if it works than i will post what i found

deviantony:
Is this really a fix though? :)

PR #326, opened 2018-10-02
Add APM server and APM test agent to compose file

HarisSistek:
Unsure if this of interest to you, but was doing some prototyping with Elasticsearch APM server using your elk stack and wanted to return the favor if an APM server is wanted in the stack. 

Why:
* Adds Elasticsearch Application Performance Monitoring
* Can view application traces, errors and alerts trough Kibana

How:
* Add apm-server container to the network which will receive traffic on port 8200 and report to elastic search server at port 9200
* Add a test flask app with redis to view apm tracing

### DIFF
--- a/extensions/apm-server/Dockerfile
+++ b/extensions/apm-server/Dockerfile
@@ -1,0 +1,3 @@
+ARG ELK_VERSION
+
+FROM docker.elastic.co/apm/apm-server:${ELK_VERSION}
--- a/extensions/apm-server/README.md
+++ b/extensions/apm-server/README.md
@@ -1,0 +1,54 @@
+# APM Server extension
+
+Adds a container for Elasticsearch APM server. Forwards caught errors and traces to Elasticsearch
+server that can be viewed in Kibana. 
+
+## Usage
+
+If you want to include the APM server, run Docker compose from the root of 
+the repository with an additional command line argument referencing the `apm-server-compose.yml` file:
+                                                                           
+```bash
+$ docker-compose -f docker-compose.yml -f extensions/apm-server/apm-server-compose.yml up
+```
+
+## Connecting an agent to APM-Server
+
+The most basic configuration to send traces to apm server. Is to specify the
+`SERVICE_NAME` and `SERVICE_URL`. Here is an example Python FLASK configuration: 
+
+```python
+import elasticapm
+from elasticapm.contrib.flask import ElasticAPM
+
+from flask import Flask
+
+app = Flask(__name__)
+app.config['ELASTIC_APM'] = {
+    # Set required service name. Allowed characters:
+    # a-z, A-Z, 0-9, -, _, and space
+    'SERVICE_NAME': 'PYTHON_FLASK_TEST_APP',
+
+    # Set custom APM Server URL (default: http://localhost:8200)
+    'SERVER_URL': 'http://apm-server:8200',
+
+    'DEBUG': True,
+}
+```
+
+More configuration setting can be found under the **Configuration**
+section for each language. Link: https://www.elastic.co/guide/en/apm/agent/index.html 
+
+## Checking Connectivity and Importing default APM Dashboards
+
+From Kibana main window press:
+
+1. `Add APM` button under Add Data to Kibana section
+2. Ignore all the install instructions and press `Check APM Server status` button.
+3. Press `Check agent status`
+4. Press `Load Kibana objects` to get the default dashboards
+5. Lastly press the `APM dashboard` to the bottom right.
+
+## APM Agent Documentation
+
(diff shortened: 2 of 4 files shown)


deviantony:
IMO should be added as an extension rather than as a part of the core, see https://github.com/deviantony/docker-elk#how-can-i-enable-the-provided-extensions

HarisSistek:
@deviantony good suggestion, much cleaner configuration this way. Moved it to two extensions, one for the apm server and one for the apm agent.

deviantony:
@HarisSistek why packaging it under two extensions instead of one?

HarisSistek:
Maybe easier to explain with this image:
image

The APM agents and beats are meant to be at hosts and send data to their respective servers. 

The `apm-agent-compose.yml` was just meant as en example flask app on how to connect an apm agent to the apm server and send traces. I could remove the apm-agent from `extensions` dir to perhaps a new dir called `example-connectors` that shows how to connect beats and apm-agents.

What do you suggest?

deviantony:
Ok, thanks for the illustrated answer. Makes sense. 

Similar to the fact that we do not have a beats extension, I think that we should remove the apm-agent extension and instead document the connection from agent to server and how to send traces in the Wiki instead (in a new "Connectors" section for example).

Thoughts @antoineco ?

HarisSistek:
@deviantony @antoineco So to recap, I will:
- remove the apm-agent
- add description on how to connect apm agents to the apm server in the extensions/apm-server/readme

About making pre-buildt image and demo app: Unsure how much time I have to make them. Also they should probably be made under the deviantony dockerhub domain so people know that they are related  to the github repo. E.g. deviantony/elk-demo-apm-agent

antoineco:
Sounds good to me!

Could we apply the agent instructions to a generic component that's already available on the Docker Hub? (bear with me, I've never used APM)

HarisSistek:
Well the apm agent has to wrap the running process and forward data to the apm-server. 

e.g. java:
```
java \
 -javaagent:/.../elastic-apm-agent-<version>.jar\
 -Delastic.apm.service_name=my-application\
 -Delastic.apm.server_url=http://localhost:8200\
 -Delastic.apm.application_packages=org.example\
 -jar my-application.jar
```

while nodejs:
```
// Add this to the VERY top of the first 
// file loaded in your app
var apm = require('elastic-apm-node').start({
  // Override service name from package.json 
  // Allowed characters: a-z, A-Z, 0-9, -, _, 
  // and space
  serviceName: '',
  
  // Use if APM Server requires a token
  secretToken: '',
  
  // Set custom APM Server URL
  // Default: http://localhost:8200
  serverUrl: ''
})
``` 

I haven't looked in dockerhub if there are any ready examples. You can read more about the elk apms here: https://www.elastic.co/solutions/apm and each language has its own apm doc that explains how to use it.

HarisSistek:
@deviantony @antoineco I removed the apm-agent, and added some more information on the apm-server/readme on connecting, finding apm agent docs etc. Do tell if you want anything changed

.gitignore
```diff
@@ -0,0 +1 @@
+.idea/*
```
antoineco: Not sure if we really need that, this project is essentially Dockerfiles and a YAML file.

extensions/apm-server/Dockerfile
```diff
@@ -0,0 +1,3 @@
+ARG ELK_VERSION
+
+FROM docker.elastic.co/apm/apm-server:${ELK_VERSION}
```
antoineco: Please make sure all committed files end with a newline character ([POSIX compliance](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206))

> Preferences > Editor > General > Ensure line feed at file end on save

.gitignore
deviantony: Agree, no need to add the .gitignore file.

extensions/apm-server/README.md
```diff
@@ -0,0 +1,54 @@
...
+    'DEBUG': True,
+}
+```
+
+More configuration setting can be found under the **Configuration**
+section for each language. Link: https://www.elastic.co/guide/en/apm/agent/index.html 
+
+## Checking Connectivity and Importing default APM Dashboards
+
+From Kibanas main window press:
```
deviantony:
```suggestion
From Kibana main window press:
```

HarisSistek:
@deviantony @antoineco thx for the feedback, here you go.

deviantony:
Thanks @HarisSistek :+1:

HarisSistek:
@deviantony @antoineco No problem, just repaying the favor. Your stack helped a lot while prototyping

PR #337, opened 2018-12-10
Write Curator logs to STDOUT

tsauerwein:
So far the included Curator configuration logs to `/usr/share/curator/curator.log`. To follow the best-practices in regards to logging inside a container, this PR proposes to log to `STDOUT`.

### DIFF
--- a/extensions/curator/config/curator.yml
+++ b/extensions/curator/config/curator.yml
@@ -20,5 +20,5 @@ client:
 
 logging:
   loglevel: DEBUG
-  logfile:  /usr/share/curator/curator.log
+  logfile:
   logformat: default


antoineco:
@tecris any thoughts on that?

shoepping:
I can only suspect "best-practices in regards to logging inside a container" is about https://docs.docker.com/config/containers/logging/. 
If that's the case and if this change honors those recommendations, IMHO this PR is in the right direction.

tecris:
I can only suspect "best-practices in regards to logging inside a container" is about https://docs.docker.com/config/containers/logging/. 
If that's the case and if this change honors those recommendations, IMHO this PR is in the right direction.

antoineco:
Sounds reasonable, thanks @tsauerwein! :raised_hands:

PR #360, opened 2019-02-06
Make ELK image flavour configurable

mw-jko:
This project was using hardcoded `-oss` images. Playing with subscription plan features was only possible when hacking the Dockerfiles. Thus, I made the flavour configurable.

While doing this I also bumped searchguard to latest versions (separate commit, so the other change can be cherrypicked)

### DIFF
--- a/.env
+++ b/.env
@@ -1,5 +1,5 @@
 ELK_VERSION=6.5.4
-# set to -oss if you prefer the open source license, leave empty otherwise
+# Leave blank to use the "basic" image flavours, which include X-Pack.
 # see https://www.elastic.co/subscriptions
 ELK_FLAVOUR=-oss
 
--- a/.env
+++ b/.env
@@ -5,5 +5,5 @@ ELK_VERSION=6.5.4
 #  * https://search.maven.org/search?q=g:com.floragunn%20AND%20a:search-guard-6&core=gav
 #  * https://search.maven.org/search?q=g:com.floragunn%20AND%20a:search-guard-kibana-plugin&core=gav
 #
-SG_VERSION=24.0
-SG_VERSION_KIBANA=17
+SG_VERSION=24.1
+SG_VERSION_KIBANA=18
--- a/.env
+++ b/.env
@@ -8,5 +8,5 @@ ELK_FLAVOUR=-oss
 #  * https://search.maven.org/search?q=g:com.floragunn%20AND%20a:search-guard-6&core=gav
 #  * https://search.maven.org/search?q=g:com.floragunn%20AND%20a:search-guard-kibana-plugin&core=gav
 #
-SG_VERSION=24.1
-SG_VERSION_KIBANA=18
+SG_VERSION=24.0
+SG_VERSION_KIBANA=17
--- a/.env
+++ b/.env
@@ -1,4 +1,7 @@
 ELK_VERSION=6.5.4
+# set to -oss if you prefer the open source license, leave empty otherwise
+# see https://www.elastic.co/subscriptions
+ELK_FLAVOUR=-oss
 
 # Check Search Guard versions matrix at https://docs.search-guard.com/latest/search-guard-versions
 # or the Maven Central repository:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,6 +6,7 @@ services:
     build:
       context: elasticsearch/
       args:
+        ELK_FLAVOUR: $ELK_FLAVOUR
         ELK_VERSION: $ELK_VERSION
         SG_VERSION: $SG_VERSION
     volumes:
@@ -22,6 +23,7 @@ services:
     build:
       context: logstash/
       args:
+        ELK_FLAVOUR: $ELK_FLAVOUR
         ELK_VERSION: $ELK_VERSION
     volumes:
       - ./logstash/config/logstash.yml:/usr/share/logstash/config/logstash.yml:ro
@@ -40,6 +42,7 @@ services:
     build:
       context: kibana/
       args:
+        ELK_FLAVOUR: $ELK_FLAVOUR
         ELK_VERSION: $ELK_VERSION
         SG_VERSION_KIBANA: $SG_VERSION_KIBANA
     volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,8 +6,8 @@ services:
     build:
       context: elasticsearch/
       args:
-        ELK_FLAVOUR: $ELK_FLAVOUR
         ELK_VERSION: $ELK_VERSION
+        ELK_FLAVOUR: $ELK_FLAVOUR
         SG_VERSION: $SG_VERSION
     volumes:
       - ./elasticsearch/config/elasticsearch.yml:/usr/share/elasticsearch/config/elasticsearch.yml:ro
@@ -23,8 +23,8 @@ services:
     build:
       context: logstash/
       args:
-        ELK_FLAVOUR: $ELK_FLAVOUR
         ELK_VERSION: $ELK_VERSION
+        ELK_FLAVOUR: $ELK_FLAVOUR
     volumes:
       - ./logstash/config/logstash.yml:/usr/share/logstash/config/logstash.yml:ro
       - ./logstash/pipeline:/usr/share/logstash/pipeline:ro
@@ -42,8 +42,8 @@ services:
     build:
       context: kibana/
       args:
-        ELK_FLAVOUR: $ELK_FLAVOUR
         ELK_VERSION: $ELK_VERSION
+        ELK_FLAVOUR: $ELK_FLAVOUR
         SG_VERSION_KIBANA: $SG_VERSION_KIBANA
     volumes:
       - ./kibana/config/:/usr/share/kibana/config:ro
--- a/elasticsearch/Dockerfile
+++ b/elasticsearch/Dockerfile
@@ -1,7 +1,7 @@
 ARG ELK_VERSION
 ARG ELK_FLAVOUR
 
-# https://github.com/elastic/elasticsearch-docker / https://www.docker.elastic.co/
+# https://github.com/elastic/elasticsearch-docker
 FROM docker.elastic.co/elasticsearch/elasticsearch${ELK_FLAVOUR}:${ELK_VERSION}
 
 COPY config/sg/ config/sg/
--- a/elasticsearch/Dockerfile
+++ b/elasticsearch/Dockerfile
@@ -1,7 +1,8 @@
 ARG ELK_VERSION
+ARG ELK_FLAVOUR
 
-# https://github.com/elastic/elasticsearch-docker
-FROM docker.elastic.co/elasticsearch/elasticsearch-oss:${ELK_VERSION}
+# https://github.com/elastic/elasticsearch-docker / https://www.docker.elastic.co/
+FROM docker.elastic.co/elasticsearch/elasticsearch${ELK_FLAVOUR}:${ELK_VERSION}
 
 COPY config/sg/ config/sg/
 COPY bin/ bin/
--- a/kibana/Dockerfile
+++ b/kibana/Dockerfile
@@ -1,7 +1,7 @@
 ARG ELK_VERSION
 ARG ELK_FLAVOUR
 
-# https://github.com/elastic/kibana-docker / https://www.docker.elastic.co/
+# https://github.com/elastic/kibana-docker
 FROM docker.elastic.co/kibana/kibana${ELK_FLAVOUR}:${ELK_VERSION}
 
 # Search Guard plugin
--- a/kibana/Dockerfile
+++ b/kibana/Dockerfile
@@ -1,7 +1,8 @@
 ARG ELK_VERSION
+ARG ELK_FLAVOUR
 
-# https://github.com/elastic/kibana-docker
-FROM docker.elastic.co/kibana/kibana-oss:${ELK_VERSION}
+# https://github.com/elastic/kibana-docker / https://www.docker.elastic.co/
+FROM docker.elastic.co/kibana/kibana${ELK_FLAVOUR}:${ELK_VERSION}
 
 # Search Guard plugin
 ARG ELK_VERSION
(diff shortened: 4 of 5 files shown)


antoineco:
Why not, I'm just curious about one thing: why would you use the non-OSS version together with Search Guard? Nowadays the only difference between the OSS and non-OSS images is the X-Pack Security plug-in, which is incompatible with Search Guard.

mw-jko:
Hm I was just looking on the [subscription overview](https://www.elastic.co/subscriptions) on the Elastic website. And to me it looked like there is quite a difference between "open source" and "basic" (which are both free and none includes authentification). 

I'm about to upgrade our stack to 6.x but tbh I was a bit in a rush and not able (yet) to check if it's compatible with searchguard. 😊 Did not think this far (about "searchguard might not work").

So feel free to close to close the PR for now and I will create a new one after testing.

antoineco:
Ah, "OSS" and "basic" subscriptions are different from the OSS / non-OSS images. You can use both with either images, it's just a commercial aspect.

You only need the non-OSS image if you chose a subscription above _Gold_, in which case you wouldn't use Search Guard, because Search Guard is an alternative to _X-Pack Security_ (included at the _Gold_ level).

If you want to try _X-Pack Security_ for free for 30 days you can check out the `x-pack` branch.

mw-jko:
Ah so this isn't required at all. I made the following assumptions:

* `-oss` = oss version stripped of the basic features as well, no license handling included
* without `-oss` and without configuring/adding a license = basic "license"
* without `-oss` and with configuring/adding a license = gold/platinum

I wonder what's the difference between `-oss`/non`-oss` *images* then 🤔

antoineco:
The only difference is the absence of X-Pack Security features, which are released under a difference license (the _Elastic license). All other X-Pack features (monitoring, alerting, machine learning) are released under the Apache 2.0 license, and are therefore considered "real" open-source. You can use these features with Search Guard out-of-the-box with the `-oss` images, for free.

You can read about both licenses here: https://github.com/elastic/elasticsearch/tree/master/licenses

mw-jko:
Ok, so we just close this PR then? :)

antoineco:
If the intention was to use X-Pack together with Search Guard this PR is not required, the `-oss` images contains all the components supported by Search Guard.

If after your installation you realize there is indeed something missing from the `-oss` images that Search Guard supports, we can revisit the PR and document the new variable. What do you think?

mw-jko:
> If after your installation you realize there is indeed something missing from the `-oss` images that Search Guard supports, we can revisit the PR and document the new variable. What do you think?

Sounds good to me. I will close the PR and re-open / create new one when needed.

mw-jko:
Hi @antoineco,

it worked fine to me until now when I wanted to do an upgrade from 6.6.2 to 6.7.0 for the 7.0 upgrade assistant. Then I realized I'm missing the "Elasticsearch" management section in Kibana where the "new" upgrade assistant can be found. 

I asked in the searchguard forums about this and the conclusion was: it does not work with `-oss` because there is no xpack included.

So I would like to reopen this and improve my pull request. 

What is your suggestion on this? Allow configuration for image-flavour is a must, disabling `xpack.security` as well to make searchguard work (I hope this setting causes no trouble with `-oss` flavour but I will check). 

In addition I would also disable all the gold+platinum xpack features.

antoineco:
Fair point, I can confirm `GET _nodes/plugins` doesn't show any X-Pack plugin like it used to (if I'm not mistaken). Trying to install the plugin manually also throws an error:
```console
$ docker-compose exec elasticsearch elasticsearch-plugin install x-pack
ERROR: X-Pack is not available with the oss distribution; to use X-Pack features use the default distribution
```

mw-jko:
I think it's an overlooking from elastic itself (I'm not sure at which point the `-oss` flavour was introduced - 6.0 or before?) but in 5.6 the upgrade wizard was available for non-xpack as well. 

I have no idea how to deal with this atm though. I will probably hack something like non`-oss` with disabled xpack.security for myself but this isn't possible for the project itself (or only with a switch away from `-oss`)

docker-compose.yml
```diff
@@ -6,6 +6,7 @@ services:
     build:
       context: elasticsearch/
       args:
+        ELK_FLAVOUR: $ELK_FLAVOUR
```
antoineco: Maybe keep `ELK_FLAVOUR` below `ELK_VERSION` to keep the order of the env file.

elasticsearch/Dockerfile
```diff
@@ -1,7 +1,8 @@
 ARG ELK_VERSION
+ARG ELK_FLAVOUR
 
-# https://github.com/elastic/elasticsearch-docker
-FROM docker.elastic.co/elasticsearch/elasticsearch-oss:${ELK_VERSION}
+# https://github.com/elastic/elasticsearch-docker / https://www.docker.elastic.co/
```
antoineco: https://www.docker.elastic.co/ is only a list of images, not so relevant in the context IMO.

.env
```diff
@@ -1,9 +1,12 @@
...
+ELK_FLAVOUR=-oss
 
 # Check Search Guard versions matrix at https://docs.search-guard.com/latest/search-guard-versions
 # or the Maven Central repository:
 #  * https://search.maven.org/search?q=g:com.floragunn%20AND%20a:search-guard-6&core=gav
 #  * https://search.maven.org/search?q=g:com.floragunn%20AND%20a:search-guard-kibana-plugin&core=gav
 #
-SG_VERSION=24.0
-SG_VERSION_KIBANA=17
+SG_VERSION=24.1
```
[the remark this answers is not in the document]
mw-jko: Fine to me, doesn't change the problem with the `xpack.security.enabled` issue with `-oss` flavour though.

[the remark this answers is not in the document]
antoineco: Sure, it just makes it easier to follow changes afterwards.

docker-compose.yml
mw-jko: changed

elasticsearch/Dockerfile
mw-jko: changed with f88c8c3b292c36d19e34f8146ef356cad3ac0738

.env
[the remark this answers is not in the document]
mw-jko: reverted commit with 0c79d820d7368380215c5cb49f00aad796479596

mw-jko:
I just got the response from elastic that it's unlikely they will include the upgrade assistent in the `-oss` version.

mw-jko:
So to me there are two options left:

1. stick with `-oss` as default, but make it configurable (this PR). question to answer: how to disable the gold/platinum features for basic flavour while making sure the container does not crash in `-oss` flavour
2. move away from `-oss` again, disable gold/platinum features by default

I would prefer option 2, but it's a philosophical question. Keep in mind this will affect the `master` branch (and maybe others) as well.

antoineco:
There is a potential feature overlap between X-Pack security and Search Guard. Besides, the average user that clones this repo to give Search Guard a try will most likely want to focus on Search Guard, and prefer to ignore extra considerations such as "what should I disable and why?".

I would consider the addition of X-Pack an opt-in customization, and I think your PR solves this pretty well 👍

antoineco:
Thanks @mw-jko! Feel free to submit other PRs if you will like improving the usability.

mw-jko:
Thanks for the merge!

> There is a potential feature overlap between X-Pack security and Search Guard. Besides, the average user that clones this repo to give Search Guard a try will most likely want to focus on Search Guard, and prefer to ignore extra considerations such as "what should I disable and why?".

The missing upgrade assistant might break things for other users (even non searchguard but the very basic `master`) as well but it really depends on the use case. If one uses this repo as a "let's tryout elastic" the missing assistant is good enough. If one uses this repo in "production" not just for tryout, the missing upgrade assistant is potentially a problem (no matter if `searchguard` or `master`).

It's a philosophical question about what problem this repo should solve and how. 

We could easily switch back to non `-oss` as the default and disable all the gold/platinum features. The basic licence is included and free of charge anyway. This way both the "tryout" and the "production" faction would benefit.

antoineco:
Makes sense. 

To be clear, the assistant is already available out-of-the-box on the `x-pack` branch of this repo. Are you suggesting we should make it the default branch and move the current `master` to a new `oss` branch, for example?

I have no issue including X-Pack by default, we just have to ensure we include a warning regarding the [license](https://github.com/deviantony/docker-elk/pull/360#issuecomment-461849792).

mw-jko:
No I'm suggesting we move away from the `-oss` image per default. 

* `master`/`searchguard`: use "basic" version, disable gold/platinum features (this includes `security`)
* `searchguard`: use "basic" version, add searchguard
* `xpack`: use "basic" version, do not disable anything
* not sure about `vagrant` and `legacy` never used/looked at them

We could also add more `-oss` branches for `master`/`searchguard` but this will increase required effort for updates even more. Or revert this logic and add `-basic` branches (whatever the excact naming would be, `-basic` is a bit silly) to not change or potentially break license for existing users.

antoineco:
> `master`: use "basic" version, disable gold/platinum features (this includes security)

The lines with the `x-pack` branch would become quite blurred as the only difference would become the value of a single option: `xpack.security.enabled` (and a handful of others, but I simplify).

This would change the purpose of the 3 branches we currently maintain, which all provide different software, and not just different combinations of options. In that case I'd rather rename `x-pack` to `master` and let the user disable security features, but maybe not everyone wants that.

mw-jko:
> The lines with the x-pack branch would become quite blurred as the only difference would become the value of a single option: xpack.security.enabled (and a handful of others, but I simplify).

You can also view as this like this as well:

* `master`: all paid features disabled, no security
* `searchguard`: all paid features disabled, added searchguard security
* `x-pack`: all features enabled

There is a clear distinction between this 3 versions. There is quite a lot of features you can not use with basic subscription (there a big features like "machine learning" and alerting) but also smaller ones like PDF reporting.

antoineco:
@deviantony what's your take on that? Shall we replace the "pure OSS" flavour with a "X-Pack with paid features disabled" flavour?

mw-jko:
> @deviantony what's your take on that? Shall we replace the "pure OSS" flavour with a "X-Pack with paid features disabled" flavour?

Keep in mind we have a new situation now:

* some time ago (when I actually started to use this project back in good ol' 5.x times 😉 ): paid versus oss 
* now: paid versus free versus oss

deviantony:
Sorry for not participating in this discussion earlier.

My view on this project has always been to provide a simple way to play with Elastic stack and not to provide a production ready solution.  Running the Elastic stack in production usually implies factors and parameters that are related to the user environment and that we can't simply support.

I think that we decided to introduce the use of `-oss` images in https://github.com/deviantony/docker-elk/pull/196

Although, it was more of a "we're doing open-source, let's use open-source images" kind of decision.

With the recent changes to Elastic and their plugin open-source policies, it is definitely a good moment to reconsider how to present this stack to the users and how to configure it.

I'm keen to make switch the default branch behavior to "paid trial" by leveraging the non-oss images and by keeping enabled all the paid features (equivalent of the `x-pack` branch) as I think it's good for new users to discover the full potential of the solution (even paid features) and that existing users should be able to opt-out of paid features to revert to the "free" version.

We also want to keep support for "free"+"searchguard". Now regarding "oss"+"searchguard"? Are the "oss" images supported by "searchguard" or not? I'm thinking about making the "oss" stack version an optional deployment instead of the default one.

Here are a few things to consider before going further on this topic:

* Can we simplify it to support all these kinds of deployments inside a single branch instead of having to maintain and support multiple branches? 

E.g. can we provide a way for the user to use `FREE`+`SEARCHGUARD`, or `PAID_TRIAL` or `FREE` after cloning the repo?

I think this is the list of options that we could provide:

* PAID_TRIAL (default mode)
* FREE
* OSS
* FREE+SEARCHGUARD
* OSS+SEARCHGUARD (if supported by searchguard)

Considering this evolution either through technical solutions or documentation.

One last point, I want to put an emphasis on the importance of the documentation for this project, for example this PR introduces a new "flag" via https://github.com/deviantony/docker-elk/pull/360/files#diff-f579cccc964135c7d644c7b2d3b0d3ecR4 that should be documented in the README.

TLDR: I'm keen to replace the "pure OSS" with the "paid trial" flavour but also want to consider a simple way to switch flavour/configurations.

WDYT?

mw-jko:
> Speaking of documentation, I want to put an emphasis on its importance in this project, for example this PR introduces a new "flag" via https://github.com/deviantony/docker-elk/pull/360/files#diff-f579cccc964135c7d644c7b2d3b0d3ecR4 that should be documented in the README IMO.

Good point, will add something to the README about this over the weekend.

About the "main discussion" (not sure if a closed PR is a good place for this, but I guess it's a bit late for switching to an issue :)):

The *current* implementation *is* OSS and it works fine with searchguard. When switching to "free" we have to disable xpack security. The whole question came on the table when I realized that elastic does not offer the (major version) upgrade wizard for the oss version anymore and I tried to find a way to solve both issues (provide a easy way to support both oss and "basic" AND configure xpack as easy as possible) and realized I have no clue how to do this: 

The problem with oss versus basic is, that elastic crashes when one tries to configure features which are not included (like trying to disable security in OSS image). This makes the `-oss` toggle (this PR here) error-prone. You have to set the flavour in the .env file *and* add lines to both kibana and elasticsearch yaml files, when you want to use the "normal" (non oss) version.

> * PAID_TRIAL (default mode)
> * FREE
> * OSS
> * FREE+SEARCHGUARD
> * OSS+SEARCHGUARD (if supported by searchguard)

How do you define "default mode"? What's included in `master`? Or set `x-pack` as the default branch? (Not sure if this is possible in GitHub but I guess so) I would not change the "definition" of an branch, though. If one relies on the master branch we should not change this branch to the the paid trial. 

Otherwise sounds good to me, but I think it's increasing the required maintenance effort quite a lot. When elastic changes things you have to adjust 5 branches, when searchguard changes things still 2. 

What about this: 

* all branches are changed to non `-oss` 
* `master` and `searchguard` have paid features disabled
* we add comments in the yml files and adjust the readme how to switch to pure `oss` (like: "set the flavour, remove/comment this and that lines in this and that files")

deviantony:
> About the "main discussion" (not sure if a closed PR is a good place for this, but I guess it's a bit late for switching to an issue :)):

Never too late, I'll open a new issue.

PR #365, opened 2019-02-25
Logstash pipelines are not being picked up.

egok:
With the following error:
[WARN ][logstash.config.source.multilocal] Ignoring the 'pipelines.yml' file because modules or command line options are specified

Following the fix:
https://github.com/elastic/logstash/issues/8420


egok:
Yes, i think this is a better idea.

Let me test with the pipelines file and make a new merge request.
I will close this MR for now.

PR #374, opened 2019-03-16
Fix user name in the command of README.md

renjinghai:
The user name should be **elastic**

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ Create an index pattern via the Kibana API:
 $ curl -XPOST -D- 'http://localhost:5601/api/saved_objects/index-pattern' \
     -H 'Content-Type: application/json' \
     -H 'kbn-version: 6.6.1' \
-    -u kibana:changeme \
+    -u elastic:changeme \
     -d '{"attributes":{"title":"logstash-*","timeFieldName":"@timestamp"}}'
 ```
 


antoineco:
The user name `kibana` is actually correct. Ref: https://www.elastic.co/guide/en/elastic-stack-overview/current/built-in-users.html

renjinghai:
> The user name `kibana` is actually correct. Ref: https://www.elastic.co/guide/en/elastic-stack-overview/current/built-in-users.html

You are right, `kibana` is the build-in user name.
After setting `kibana` password, the curl command works. Ref: https://www.elastic.co/guide/en/x-pack/current/setting-up-authentication.html#set-built-in-user-passwords

Is there any better way to avoid this step?

antoineco:
The index pattern creation can be achieved through the web UI as well. It is not mandatory but Kibana prompts the user to create one after the first login.

The password change is also not strictly mandatory but recommended. The default is "changeme" for all built-in users (see Compose file and README).

renjinghai:
My issue is the curl index pattern creation command doesn't work.

Setup
```
git clone --branch x-pack https://github.com/deviantony/docker-elk.git
cd docker-elk
docker-compose up -d
```

Use **kibana** as the user, get Unauthorized failure.
```
curl -XPOST -D- 'http://localhost:5601/api/saved_objects/index-pattern' \
    -H 'Content-Type: application/json' \
    -H 'kbn-version: 6.6.1' \
    -u kibana:changeme \
    -d '{"attributes":{"title":"logstash-*","timeFieldName":"@timestamp"}}'
```
```
HTTP/1.1 401 Unauthorized
kbn-name: kibana
kbn-xpack-sig: d241b025aef9aa2f3bddbe9e4012c305
content-type: application/json; charset=utf-8
cache-control: no-cache
content-length: 204
connection: close
Date: Tue, 19 Mar 2019 04:07:15 GMT

{"statusCode":401,"error":"Unauthorized","message":"[security_exception] failed to authenticate user [kibana], with { header={ WWW-Authenticate=\"Basic realm=\\\"security\\\" charset=\\\"UTF-8\\\"\" }}"
```

Use **elastic** as the user. Works.
```
curl -XPOST -D- 'http://localhost:5601/api/saved_objects/index-pattern' \
    -H 'Content-Type: application/json' \
    -H 'kbn-version: 6.6.1' \
    -u elastic:changeme \
    -d '{"attributes":{"title":"logstash-*","timeFieldName":"@timestamp"}}'
```
```
HTTP/1.1 200 OK
kbn-name: kibana
kbn-xpack-sig: d241b025aef9aa2f3bddbe9e4012c305
content-type: application/json; charset=utf-8
cache-control: no-cache
content-length: 230
connection: close
Date: Tue, 19 Mar 2019 04:11:36 GMT

{"type":"index-pattern","id":"15c757d0-49fd-11e9-bdce-95fac7a87368","attributes":{"title":"logstash-*","timeFieldName":"@timestamp"},"migrationVersion":{"index-pattern":"6.5.0"},"updated_at":"2019-03-19T04:11:35.244Z","version":1}
```

antoineco:
_Sorry for the response time!_

You're absolutely right, the bootstrap password does not apply to built-in users anymore (the procedure to set these is described [here](https://www.elastic.co/guide/en/x-pack/current/setting-up-authentication.html#set-built-in-user-passwords)).

Let's keep it simple and use `elastic` indeed. Nice catch!

PR #383, opened 2019-04-15
Updated the logstash.yml to run on the docker-composer up command

willsoares:
Without this changes, the logstash can't run properly, returning 401 when running the `docker-compose up`.
This pull request changes the logstash.yml configuration to use the `elastic` user with the `changeme` password

### DIFF
--- a/logstash/config/logstash.yml
+++ b/logstash/config/logstash.yml
@@ -7,6 +7,5 @@ xpack.monitoring.elasticsearch.url: http://elasticsearch:9200
 
 ## X-Pack security credentials
 #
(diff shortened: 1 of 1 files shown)


willsoares:
Ok. But we need to mention this on the README.
I'll do that later

antoineco:
Yes that's a good suggestion :+1:  it would shed some lights on those built-in users most newcomers probably don't know about.

antoineco:
@willsoares I ended up accepting your suggestion, we want to provide users with a simple setup. I included your changes in #387 and made you a co-author.

PR #396, opened 2019-05-24
updating xpack credentials to v7.X.X

chrisbloemker:
v7 requires slightly different syntax of the xpack security credentials to connect. `elasticsearch.username` and `elasticsearch.password` results in kibana errors not being able to authenticate. Kibana webpage continues to say "Kibana server is not ready yet" while producing the auth errors. Tested on Kibana 7.0.1

### DIFF
--- a/kibana/config/kibana.yml
+++ b/kibana/config/kibana.yml
@@ -9,5 +9,5 @@ xpack.monitoring.ui.container.elasticsearch.enabled: true
 
 ## X-Pack security credentials
 #
-elasticsearch.username: elastic
-elasticsearch.password: changeme
+xpack.monitoring.elasticsearch.username: elastic
+xpack.monitoring.elasticsearch.password: changeme
--- a/kibana/config/kibana.yml
+++ b/kibana/config/kibana.yml
@@ -9,5 +9,7 @@ xpack.monitoring.ui.container.elasticsearch.enabled: true
 
 ## X-Pack security credentials
 #
+elasticsearch.username: elastic
+elasticsearch.password: changeme
 xpack.monitoring.elasticsearch.username: elastic
 xpack.monitoring.elasticsearch.password: changeme


antoineco:
Thanks for the suggestion, however this makes tests fail with
`"statusCode":401,"response":"{\"error\":{\"root_cause\":[{\"type\\\":\"security_exception\",\"reason\":\"missing authentication credentials for REST request [/_xpack]\"`

kibana/config/kibana.yml
```diff
@@ -9,5 +9,5 @@ xpack.monitoring.ui.container.elasticsearch.enabled: true
 
 ## X-Pack security credentials
 #
-elasticsearch.username: elastic
-elasticsearch.password: changeme
+xpack.monitoring.elasticsearch.username: elastic
```
antoineco: X-Pack monitoring and X-Pack security are 2 different things.

chrisbloemker: Hmm.. okay If you try to spin up the 7.0.1 stack with the former, I get the error you posted above, is that to do with the `sed` steps in travis?

antoineco: Probably yes, but that's only part of the problem. The options we use right now are perfectly valid, the documentation is clear about it: https://www.elastic.co/guide/en/kibana/current/using-kibana-with-security.html.

Did you maybe want to _add_ those monitoring options to the existing ones?

antoineco: OK, found my answer in https://www.elastic.co/guide/en/kibana/current/monitoring-data.html

> If the Elastic security features are enabled on the monitoring cluster, you must provide a user ID and password so Kibana can retrieve the data. ...

chrisbloemker: I understand what happened now, during my testing I had appended those extra xpack entries instead of replacing them. I had assumed originally that having both sets would conflict, but it seems that is not the case. Updated the PR with the latest commit.

antoineco:
Looking good, thanks @chrisbloemker! 🙌

antoineco:
Reverted. See docs: https://www.elastic.co/guide/en/kibana/7.0/monitoring-data.html

> If these settings are omitted, Kibana uses the `elasticsearch.username` and `elasticsearch.password` setting values.

chrisbloemker:
Ahh, darn. :/  I guess I'll just use that for a local workaround for now.

antoineco:
No worries :)
What problem are you facing exactly? I can't see any warning or even error on my local setup.

cbloemker:
For now I just disabled the trial, and went back to using `elasticsearch.username` and `elasticsearch.password` in my kibana.yml. I was getting lots of auth errors trying to run on a single node swarm where ES errors were preventing Kibana from loading. I think I have it stable for now. If I find anything funky, I'll let you know :)

antoineco:
Alright thanks for the feedback! Are you using the stack file from the repo to run the stack on Swarm?
If yes, did you maybe upgrade from v6.7 recently?

chrisbloemker:
Sure thing! No upgrade, this was a fresh installation on a new host, brand new docker swarm. Yes, I was using the `docker-stack.yml` for deploying to a single node swarm.

PR #414, opened 2019-07-09
Fixing the apm configuration

YosBD:
I could not connect to the apm server with nodeJS until I made this change.

I found the solution in here https://github.com/elastic/apm-agent-nodejs/issues/256#issuecomment-367944675


antoineco:
Makes sense, this probably never worked with `localhost` actually. Thanks!

PR #459, opened 2019-12-19
Update Dockerfile to use latest curator

codingsteff:
Otherwise it fails with this exception:
`elasticsearch.exceptions.ElasticsearchException: Unable to create client connection to Elasticsearch.  Error: Elasticsearch version 7.4.1 incompatible with this version of Curator (5.5.4)`


antoineco:
Thank you! :raised_hands_

PR #463, opened 2020-01-16
Open udp logstash port 5000  for logspout

DoubleDi:
Hey! Thanks for your great repo on elk!
Had an issue when installing the logspout extension. We use udp to send data to logstash, but the port is opened only for tcp connections. This PR fixes that

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -41,7 +41,8 @@ services:
         target: /usr/share/logstash/pipeline
         read_only: true
     ports:
-      - "5000:5000"
+      - "5000:5000/tcp"
+      - "5000:5000/udp"
       - "9600:9600"
     environment:
       LS_JAVA_OPTS: "-Xmx256m -Xms256m"


antoineco:
Hey, thanks for the contribution! This only applies to ports exposed on the host, containers that you start via compose are in the same network and are not affected by those settings. May I ask how you started Logspout?

antoineco:
ping @DoubleDi

DoubleDi:
Yes, i cloned this repo to 2 different machines. 
One had containers, which logs i wanted to see on kibana - i started only logspout there.
Other was the main machine - i started everithing else there. 
So i think that happened because i had 2 machines and i was sending logs between them.

antoineco:
In that case the change is necessary indeed, thanks for the clarification!

The change looks sensible because several Logstash input plugins work over UDP, so this could save users some unnecessary headaches. Thanks for the contribution!

Issue #467, opened 2020-02-25
Add historical version branches of ELK


m2kar:
### Problem description
In some cases, developers especially new starter need to use a historical version of ELK for testing and learning. As can be found in issues, there are some historical issues and pull requests about versions upgrade. 

Could you please open some branches or set some tags for historical versions. Especially for 6.x and 5.x. 

#### some history issue about version upgrade
1. https://github.com/deviantony/docker-elk/pull/382
2. https://github.com/deviantony/docker-elk/pull/362
3. https://github.com/deviantony/docker-elk/pull/339
4. https://github.com/deviantony/docker-elk/pull/290
5. https://github.com/deviantony/docker-elk/pull/286

antoineco:
Maintaining one branch per _major_ version shouldn't be an issue, since those do not get updated that often. E.g. `release-5.x`, `release-6.x` and later `release-7.x` once 8.x makes it to `master`.

For a more granular version selection, users can use the `.env` file after switching to the desired release branch.

Would that work @m2kar?

antoineco:
@deviantony I'll take that thumb up as an endorsement and send 2 PRs your way so we can discuss over there 🙂

deviantony:
Will review these 2 PRs this week-end !

deviantony:
So I was looking at these 2 relateds PRs... The 5-x version is not working and while I was troubleshooting the configuration I stumbled upon the Elastic EOL policy.

While I agreed to have major release branch versions in this project, I wonder if we need to support all the major ones? I don't see the point of maintaining/creating a branch for a software stack that is declared EOL.

https://www.elastic.co/support/eol

Thoughts @antoineco @m2kar ?

antoineco:
I agree that EOL versions are quite a gray area. What I can understand, however, is that there are still more than one ES5 instances in the wild, and that developers/operators may appreciate using this stack as a quick go-to.

To keep everyone happy, I can suggest we add a warning at the top of the 5.x README mentioning the branch will only exist until the release of 8.x. What do you think?

@m2kar do you personally use 5.x?

deviantony:
My thoughts are: we should not bother with 5.x. 

After all, this project is supposed to help you get started with the Elastic stack, I don't believe there are that many users out there using EOL versions of the stack. 

And in that case, if they want to use 5.x or any other EOL version, they'll have to support by configuring the repo themselves.

@m2kar I'm quite interested about that feedback around 5.x too.

[...]

PR #469, opened 2020-02-28
Create release-6.x branch

antoineco:
Create a branch for pinning the stack version to `6.x`, as requested in #467

Changes between `6.x` and `current`:

none! :tada:

cc @m2kar


deviantony:
I'll let you merge that one, we should probably discuss this in the related issue but it'd be good to have a notice in the README about this.

antoineco:
I'll definitely add a notice once the plan for 5.x is clarified. Thanks for the review!

PR #470, opened 2020-02-29
Add app-search

jwsy:
I got app-search running on top of your awesome work. 

I'm not sure how you want to incorporate this to your baseline, perhaps as another repo or a branch (elka?), but hope I can save you a little time the way you've saved me a ton of time.

Changes of note: 
* Update ELK version to 7.6.0
* Update README to reflect how app-search wants more mem (now asking for 5g total)
* elasticsearch.yml
  * Add additional security configs 
  * Set license to default to basic


antoineco:
@jwsy that's a great example of integration, and exactly what this stack was meant for :+1:

To me [App Search](https://www.elastic.co/app-search) falls into the category of "External applications" described in the wiki: https://github.com/deviantony/docker-elk/wiki/External-applications. We already have a small amount of them in the `extensions` directory of the repo. The APM Server is a great example of this: it's another Elastic product meant to be used together with the Elastic stack, so it totally makes sense as an extension.

The problem we currently face with extensions is that neither @deviantony nor myself maintain them, so they occasionally break. To accept this as an `extension`, I would expect a few things to ensure some quality standards are met:

* It is shipped as a separate Compose file meant to be called with `docker-compose -f docker-compose.yml -f extensions/app-search/compose.yml`, so that it becomes an _opt-in_ component.
* If any change to the configuration of the core ELK components is required to run the extension, it should be briefly documented in a README in the `extensions/app-search` directory, instead of enforced on the user (referencing a page in the Elastic documentation is fine too, cf. the apm-server extension already mentioned).
* It has a small test, [similar to our current tests](https://github.com/deviantony/docker-elk/blob/master/.travis/run-tests.sh#L60-L999), that we can execute in the CI to ensure the component is working as expected. It can be as simple as ensuring the server is up and running and sending a few requests to it to assert it is properly paired with Elasticsearch.

Does this sound doable for you?

jwsy:
Ha, this project had such good default configs that I didn't even think of the extensions - my bad for not reading all of the docs!

I'll try to submit though it may take a while since this is a side project (and like most other submitters I'm unlikely to maintain it). Either way, hope it's helpful to someone looking to use App Search on a solid ELK stack.

antoineco:
Great
Most of the work seems already done, so whenever you want to revisit this, feel free to reopen.

PR #471, opened 2020-03-02
Add App Search extension

jwsy:
Ref #470 - this is a PR to add App Search as an extension as recommended by @antoineco (great instructions, thank you)!

I don't know how you'd like to have me add test cases, but a straightforward one seems to use the given App Search credentials passed as environment variables in the compose yml. I've added those instructions to the README as well as a note about how to update root directory docker-compose.yml to use the Elasticsearch configs included in this extension.

```bash
# PASS 
$ curl 'http://localhost:3002/as/' -u app_search:changeme
<!DOCTYPE html5><html lang="en"><head><title>App Search</title><meta name="csrf-param" content="authenticity_token" />
<meta name="csrf-token" content="zlqJdX5Jc9ZSqhVnxanIJsB8s6bx9cYzvVb/1u6yokboJGALvIXbRwb/IsOYJJkZ3NePNUf/VYd245D6shE1qA=="
... Lots more data

# FAIL
$ curl 'http://localhost:3002/as/' -u app_search:WRONG_PASSWORD
<html><body>You are being <a href="http://localhost:3002/login">redirected</a>.</body></html>%```

### CODE REVIEW

File: extensions/app-search/config/elasticsearch.yml [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+xpack.license.self_generated.type: basic
+xpack.monitoring.collection.enabled: true
+xpack.security.enabled: true
+xpack:
+    security:
+      authc:
+        realms:
+          native:
+            native1:
+              order: 0
```
antoineco: Since the required configuration changes are moderate (license, realm and create_index if I understood correctly?), I would prefer to see them documented in the README instead of in a separate file.

Two reasons for that:

* Users might want to run other extensions in parallel, in which case we encourage them to consciously do the changes required for their own needs.
* If we add or change some of our defaults in the future, we will also have to juggle between multiple config files and perform the changes in multiple locations.

A code snippet in your README should be enough.

Example:

> This extension requires you to perform the following changes to the Elasticsearch configuration file:
> * Use the basic license
> * Allow the automation creation of App Search's indexes
> * Configure the authentication realms

```yaml
action.auto_create_index: ". app-search-*-logs-*,-.app-search-*,+*"

xpack.license.self_generated.type: basic

xpack:
  security:
    # ...
```

extensions/app-search/README.md
```diff
@@ -0,0 +1,57 @@
+# App Search Server extension
+
+Adds a container for Elasticsearch App Search. 
+
+## Usage
+
+To include App Search:
+1. Update the root Docker Compose file to use the elasticsearch config in this extensions folder. **Note this requires you to use the basic license as of 7.6.0**
```
antoineco: I've been wondering why we need the users to switch their license type for App Search. As far as I know, `basic` allows _less_ features. Isn't it something we would want to let users decide once they reach the end of the trial, as described in the main README?

extensions/app-search/config/app-search.yml
```diff
@@ -0,0 +1,165 @@
...
+# Set key to persist user sessions through process restarts.
+#
+#secret_session_key:
+#
+# ----------------------------- Diagnostics report ----------------------------
+#
+# Path where diagnostic reports will be generated.
+#
+#diagnostic_report_directory: diagnostics
+#
```
antoineco: If this block of commented options was copied from somewhere, could we just point users to the place they could find it instead? (e.g. the App Search repo, or just the official docs) Otherwise some things will change in new versions and become irrelevant.

antoineco:
One more comment: I found an [example of configuration](https://swiftype.com/documentation/app-search/self-managed/installation#docker) for Docker Compose in App Search's official docs, and this example is functional with a very minimal Elasticsearch configuration.

If the configuration of authentication realms and indexes is optional, I'd rather stick to the defaults and let users find tune their setup.

All services in the stack should be as unopinionated as possible to guarantee a neutral starting point for everyone.

extensions/app-search/config/elasticsearch.yml
antoineco: Apparently these settings are not even required for App Search to work (see my comment in the main thread)

jwsy: I put these in here so you'd have an app_search user in a consistent experience to the other apps, but have removed them.

extensions/app-search/README.md
jwsy: Yeah, odd that giving *less* features is required but I just wanted to get it to work quickly (for self) and am not going to read too far into it until the time comes...

jwsy:
Updates made, thank you very much for your great examples & guidance @antoineco !

antoineco:
@jwsy Thank **you** for the good work!

I'll push a commit or two to your PR in order to include some tests that we can run in Travis CI, and then we should be ready to merge.

antoineco:
@jwsy the App Search tests [are green](https://travis-ci.org/deviantony/docker-elk/builds/658451741#L701-L774). I also made some cosmetic changes to the README so that it's more aligned with the main documentation.

I'll merge as soon as you confirm you're satisfied with the current state.

jwsy:
More than satisfied with the current state (excellent improvement to the docs) and thank you very much for being a great example, @antoineco! 

I appreciate the excellent direction you've provided me, and thank you for being a wonderful steward for this project that has been very helpful to many.

antoineco:
🙌

PR #492, opened 2020-05-03
Added restart: unless-stopped to the docker-compose.yml

HristoKolev:
In order to start after server reboot. :)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,6 +6,7 @@ services:
       context: elasticsearch/
       args:
         ELK_VERSION: $ELK_VERSION
+    restart: unless-stopped
     volumes:
       - type: bind
         source: ./elasticsearch/config/elasticsearch.yml
@@ -31,6 +32,7 @@ services:
       context: logstash/
       args:
         ELK_VERSION: $ELK_VERSION
+    restart: unless-stopped
     volumes:
       - type: bind
         source: ./logstash/config/logstash.yml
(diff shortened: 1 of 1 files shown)


antoineco:
@HristoKolev thanks for the contribution!

I am not certain this would be a sensible default. Personally, I wouldn't want random Compose stacks to restart on my machine without me knowing. If I _do_ want to enable it, I add it to my own local copy or fork, without forcing all users to adopt my own preferences.

@deviantony opinions on that one?

HristoKolev:
Well, in my opinion, when I `docker-compose up -d` something I expect it to stay up until I say so.

Maybe that's just me. :)

HristoKolev:
I didn't realize how famous this repo is. If nobody has raised this issue up until now , then I must be in the wrong.

antoineco:
That holds true if you run the stack in a server setup, where it needs to be available all the time.

Now imagine using this locally, for backing a development flow that uses the Elastic stack. You will probably want your containers to be running when you're actively working on that project, but when you're not, I'm sure what you'd prefer giving back 2 GB of RAM to your workstation for other purposes.

Like you said the project is popular, so we need to keep it neutral by default, but easily extensible.

deviantony:
> If I do want to enable it, I add it to my own local copy or fork, without forcing all users to adopt my own preferences.

Agree, I do not consider this one as a sensible default either.

PR #495, opened 2020-05-09
enable TLS

wych42:
This pull request is for issue #402 by providing a guide doc(self-signed-tls.md) and an tls-enabled docker-compose file.

This was originated from my setup repo for dev&production which maintains the similar file structure as this repo which I found days before in friend's blog post. I am migrating my setup repo into this one and planning write a tutorial about how to setup a multi-nodes cluster with tls enabled using this repo.

As English is not my first language, the `self-signed-tls.md` may(should) contains many grammar errors, feel free to correct them.


deviantony:
@wych42 awesome contribution @wych42 ! Will try to review this soon.

antoineco:
Thanks @wych42, good job! I'll also take a look, and we need to decide whether we want to keep configs in the repo (at the risk of having to keep things commented here and there) or summarize this in a wiki page.

wych42:
@deviantony  @antoineco  Thank you. I also will keep following this.

absk1317:
This is awesome @	wych42 . Tried, and it works like a charm! 👍 💯

antoineco:
@wych42 I just wanted to let you know I was working on some instructions based on your PR, sorry for the long delay!

k8s8:
Thank you @wych42 I can't wait to try this out.
I assume once I enable TLS with the Basic version. I will be able to try the Alerting capabilities (https://www.elastic.co/guide/en/kibana/7.x/alerting-getting-started.html).

antoineco:
@k8s8 I have a setup available on the [`tls` branch](https://github.com/deviantony/docker-elk/tree/tls) based on this PR with some very rough instruction to generate suitable certificates inside `tls/README.md` (you have to navigate to the branch to see it).

After generating the certs, all you have to do is un-comment the relevant options inside all 4 config files (`elasticsearch.yaml`, `kibana.yaml`, `logstash.yaml`, `logstash.conf`).

Some early feedback would be very welcome!

antoineco:
@deviantony the more I think about it, the more I find the idea of including pre-generated certificates inside the `tls/` directory compelling. I was initially against it, but in the meantime I played a lot with TLS, and the setup isn't plug-and-play despite Elastic's tooling.

My argumentation in favor of it is:
* Our non-TLS setup is already "easy but insecure" (password: `changeme`) so people can experiment immediately, but are prompted to secure the stack early in the README.
* We already have a demo cert in the Searchguard flavor of the stack, for the same reason.

We could have certs with a "docker-elk" CN and 10-year validity that people can override by following instructions from the README, exactly like they currently do with the bootstrap password.

Just a few thoughts, what is your opinion?

deviantony:
@antoineco 

As long as we have a pretty obvious warning about the security of the stack when cloning it, I'm ok with it. This must be easy to see in the README though.

Do you want to provide an alternate way to start the stack with TLS? Aka another compose file? I think that having these instructions on how to override/upgrade your TLS setup inside another "guide file" (as it is done inside this PR) is great, would just need a reference to it in the README.

Sorry I haven't been able to review this one.

antoineco:
@deviantony An alternate Compose file wouldn't be sufficient unfortunately, because not all options can be set via the Compose file. The Logstash pipelines for example, which do require TLS to be enabled on the client if TLS is enforced in Elasticsearch.

The draft branch I created has a list of commented options instead, that the user can un-comment in each config file to enable TLS within the stack.

Example (Kibana conf):
```yaml
##
## TLS configuration
## See instructions from README to enable.
##

## Communications between Kibana and Elasticsearch
## see https://www.elastic.co/guide/en/kibana/current/configuring-tls.html#configuring-tls-kib-es
#
#elasticsearch.ssl.certificateAuthorities: [ "config/elasticsearch-ca.pem" ]
```

In the file above, the user would un-comment `elasticsearch.ssl.certificateAuthorities`, and Kibana would "just work". Why? Because all certs are already _mounted by default in the Compose file to their expected locations_. If they don't exist, fine, the stack starts anyway due to the way Docker handles non-existing mounts. Elastic components also start just fine since all TLS options are off by default.

What I explain above works out-of-the-box providing that we pre-generate all TLS certificates. If we don't, users needs to go through a series of steps to generate them, even if they just want a quick and dirty TLS-enabled playground, which could be tedious. Eventually they _should_ do that anyway, and like you said this needs to be visible right away in the README, but I think having a "it just works" alternative would also be comfortable.

deviantony:
@antoineco what about having an alternate compose file mapped to specific TLS related config files? e.g. that compose file would use `elasticsearch/config/elasticsearch.tls.yml`, ` kibana/config/kibana.tls.yml` etc...

The negative point about this one is that you need to duplicate some configuration and could be a bit painful for us to maintain but on the other side you provide an easy way to start up the stack with TLS without having to uncomment/change protocol in specific config files.

antoineco:
True, we could also maintain multiple versions of config files, I didn't think about that.

RaminMT:
Hi,
What if certificates be generated automatically during first-time launch?

antoineco:
@RaminMT that would be ideal, but it's a little bit tricky to orchestrate with Compose, because there is no notion of startup hook to execute some custom logic before starting containers.

RaminMT:
Even if we use an script file as entrypoint?
We can have a bash script which handles this tls things and also starting process and set that as entrypoint of the container (probably ElasticSearch container)

antoineco:
It's possible as long as we ensure Elasticsearch's startup script (the one that already ships with the image) is executed with all the arguments that a user may pass on the command line (`args: [...]`).

What I don't like with that approach is:

- We introduce another component on top of official images (the "script"), and I like the way we currently rely on vanilla images, without any _docker-elk_ specific integration.
- Users have no control at all over the way certificates are generated. They start an image and something, somewhere, generates critical security files. We will never satisfy everyone with this: what about domains, expiration times, passphrases?

IMO a minor amount of setup ensures users fully understand what they do, and if people want automation, they can plug it _on top_, instead of having to fork the project to replace our opinionated bits.

antoineco:
The outcome of this story can be found on the [`tls` branch](https://github.com/deviantony/docker-elk/tree/tls). **This branch only exists for demonstration purposes.**

In a nutshell:

* TLS enabled by default in Elasticsearch (HTTP + transport).
* All components (including extensions) pre-configured to communicate with Elasticsearch over HTTPS.
* Pre-generated X.509 certificates and keys for a hassle-free experience in local environments.
* Big warning in the README that prompts users to re-generate all certificates and keys, with detailed instructions.

The servers exposed by Kibana and Enterprise Search don't have HTTPS enabled because self-signed certificates cause web browsers to throw warnings and block web pages.

Closing until a better alternative is agreed upon.

PR #522, opened 2020-08-12
TLS setup

antoineco:
Revisiting #495 with a few tweaks to make the integration tighter with the current stack, and the addition of user-facing instructions.

**DRAFT**

### DIFF
--- a/kibana/config/kibana.yml
+++ b/kibana/config/kibana.yml
@@ -4,10 +4,20 @@
 #
 server.name: kibana
 server.host: 0.0.0.0
-elasticsearch.hosts: [ "http://elasticsearch:9200" ]
+elasticsearch.hosts: [ "http://elasticsearch:9200" ]  # (!) set protocol to https when TLS is enabled
 monitoring.ui.container.elasticsearch.enabled: true
 
 ## X-Pack security credentials
 #
 elasticsearch.username: elastic
 elasticsearch.password: changeme
+
+##
+## TLS configuration
+## See instructions from README to enable.
+##
+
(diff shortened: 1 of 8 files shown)


kibana/config/kibana.yml
```diff
@@ -4,10 +4,20 @@
 #
 server.name: kibana
 server.host: 0.0.0.0
-elasticsearch.hosts: [ "http://elasticsearch:9200" ]
+elasticsearch.hosts: [ "http://elasticsearch:9200" ]  # (!) set protocol to https when TLS is enabled
```
antoineco, line 7: @deviantony to your comment in https://github.com/deviantony/docker-elk/pull/495#issuecomment-673165902, multiplying configurations to support multiple variants turns out to result in a confusing experience, especially with the multiple pre-configured extensions we support.

Here's a crazy idea for you: would it be bad if we un-commented all the commented lines from this PR, and just enabled TLS by default? 🙃 It's 2020, TLS is a default almost everywhere, and addressing ES over `https` is not a significant change from a user's perspective.

deviantony: Well on the user experience side that would be great but that would open up to the concerns you stated in the following comments:

* https://github.com/deviantony/docker-elk/issues/402#issuecomment-571152166 (point 1)
* https://github.com/deviantony/docker-elk/pull/495#issuecomment-706693683 (point 2)

antoineco: Alright fair enough, I think I've run out of ideas then. I'll close that story for now because I can't really dedicate time to it.

PR #387, opened 2019-04-26
Enable Stack Features (X-Pack) and trial by default

antoineco:
We're now "secure by default" with X-Pack and paid features enabled by default 🔐 (30-day trial) 

If this gets merged, the `x-pack` branch will become the new `master` in a second step.

Closes #163
Closes #378

PR #526, opened 2020-08-28
Disabled paid features of Elasticsearch

Abs0lutZero:
I think its better to have the basic licence enabled by default instead of the other way around.

### DIFF
--- a/elasticsearch/config/elasticsearch.yml
+++ b/elasticsearch/config/elasticsearch.yml
@@ -8,6 +8,6 @@ network.host: 0.0.0.0
 ## X-Pack settings
 ## see https://www.elastic.co/guide/en/elasticsearch/reference/current/setup-xpack.html
 #
-xpack.license.self_generated.type: trial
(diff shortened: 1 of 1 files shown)


antoineco:
@Abs0lutZero thanks for the suggestion, but the usage of the trial license as a default was a conscious decision implemented in #387.

The rational is that we don't want to restrict the list of features new users can experiment with, but rather give them a better understanding of what the Elastic stack can actually offer.

Before that we had 2 flavours in different branches - Commercial and OSS - but that was tedious to maintain, so we figured `trial` was sane enough to start with since it could be disabled super easily.

Abs0lutZero:
@antoineco 

Thank you for the clarification. I had not though of it from that point of view

antoineco:
@Abs0lutZero of course the opposite standpoint would also perfectly make sense (disable paid features and show people how to enable them) but we had to come up with a decision when merging the two branches, and that's what emerged from the discussion.

Issue #455, opened 2019-12-05
Discovery not working in swarm when fresh install of 19.03.5


eoli3n:
### Problem description

Configuring elasticsearch in swarm mode as wiki says results in discovery issue when installing docker 19.03.5 on Debian Buster from scratch.

Installing 19.03.3 and run stack works
Then upgrade to 19.03.5, works too.

(did it 10 times)

### Extra information

```
# Install docker
apt update && apt-get install apt-transport-https ca-certificates curl gnupg2 software-properties-common -y
curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add -
add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/debian $(lsb_release -cs) stable"
```

#### 19.03.5 
```
apt update && apt-get install docker-ce docker-ce-cli containerd.io -y
```

#### 19.03.3
```
apt update && apt-get install docker-ce=5:19.03.3~3-0~debian-buster docker-ce-cli=5:19.03.3~3-0~debian-buster containerd.io -y
```
```
# Swarm init
docker swarm init

# Sysctl config
sysctl vm.max_map_count=262144

# Project clone and run stack
git clone https://github.com/eoli3n/docker-elk
cd docker-elk
docker stack deploy -c docker-stack.yml elk
docker service logs elk_elasticsearch -f
```

#### Stack configuration

Here diff i applied to configure docker-elk
[https://github.com/deviantony/docker-elk/compare/master...eoli3n:master](https://github.com/deviantony/docker-elk/compare/master...eoli3n:master)

##### 19.03.5 fresh
[docker service logs elk_elasticsearch](http://ix.io/23E5)

##### 19.03.3 fresh
[docker service logs elk_elasticsearch](http://ix.io/23Ee)
[curl -D- 'http://127.0.0.1:30000/_cluster/health?pretty' -u elastic:changeme](http://ix.io/23Ef)

antoineco:
For some reason you have that weird behavior where nodes discover themselves on the 10.0.0.0/24 network where the DNS resolves to 10.0.1.0/24, but I'm honestly not able to explain it. I doubt it's related to docker-elk specifically.

> {**10.0.0.8:9300**}{dilm}{ml.machine_memory=4092379136, xpack.installed=true, ml.max_open_jobs=20}]; discovery will continue using [**10.0.1.8:9300, 10.0.1.9:9300, 10.0.1.10:9300**

antoineco:
[...]

PR #542, opened 2020-10-13
Move CI flow from Travis CI to GitHub Actions

antoineco:
GitHub now has its own CI called _GitHub actions_ embedded into code repositories, and after playing with it a little bit I have to say I really like it! It's fast and quite similar to Travis, with a large catalog of "actions" and some neat integrations with Git repos.

This PR moves the CI pipeline from Travis CI to GitHub Actions, without changing the existing test logic. I had to rewrite the util functions though, to make things more portable and avoid assuming everything listens on `localhost`.

All runs, including the ones triggered by the PR, can be seen by clicking the `Actions` tab in the top bar ([here](https://github.com/deviantony/docker-elk/actions)).


antoineco:
Marking as draft until I figure why `expect` fails on `bin/elasticsearch-setup-passwords` with code `78` in Swarm mode.

antoineco:
Fixed: https://github.com/deviantony/docker-elk/issues/455#issuecomment-708035654

deviantony:
@antoineco I see I'm assigned to this one, do you want me to review it? Overall I agree with the use of Github actions as well.

antoineco:
@deviantony it was mostly for visibility, so that you can disable the Travis integration in the repo if you want.

The current run should pass, and then we should be good to merge.

deviantony:
Allgood I'll get rid of the Travis integration after this PR is merged.

Great job :+1:

antoineco:
Thanks for the quick check! 🙌

deviantony:
I removed the Travis webhook, I think it's the only thing that was setup re Travis (no integration).

PR #544, opened 2020-10-15
Update to v8.0.0

### LINKED ISSUE #543: Update to v8.0.0
Opening this issue for visibility.

Elastic v8.0.0 is right at the door and has some snapshot images released, so we can test it early before it comes GA.

---

Changelog (not yet final)

* [Elasticsearch 8.0.0 Release Notes](https://www.elastic.co/guide/en/elasticsearch/reference/master/release-notes-8.0.0-alpha1.html)
* [Kibana 8.0.0 Release Notes](https://www.elastic.co/guide/en/kibana/master/release-notes-8.0.0-alpha1.html)
* [Logstash 8.0.0 Release Notes](https://www.elastic.co/guide/en/logstash/master/releasenotes.html)

antoineco:
Sending this PR early so people can experiment with the new v8.x series.
Marking as draft until v8.0.0 gets officially released (2022-02-08 according to the [Elastic product end of life dates](https://www.elastic.co/support/eol) page).

---

Closes #543

List of changes impacting docker-elk:

### Logstash

- **The output to Elasticsearch is handled as a data stream.**

  Starting with v8.0.0, the `elasticsearch` output for Logstash sends log data to a data stream instead of `logstash-*` indices by default. The name of the default data stream is `logs-generic-default`. docker-elk remains unopinionated and simply uses Elastic's defaults like it always has, so users who prefer to retain the old behaviour need to explicitly opt-out of data streams in their Logstash pipelines.

  Refs:
  - https://www.elastic.co/guide/en/elasticsearch/reference/current/data-streams.html
  - https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-data-streams

- **The (legacy) monitoring data collection is now disabled.**

  This feature was deprecated since v7.9.0, and removed in v8.0.0.

  Ref: https://www.elastic.co/guide/en/logstash/current/monitoring-internal-collection-legacy.html

### Elasticsearch

- **The command line tool `elasticsearch-setup-passwords` was deprecated in favour of a new `elasticsearch-reset-password` tool.**

  Passwords for built-in users must now be generated one by one.

  Ref: https://www.elastic.co/guide/en/elasticsearch/reference/current/setup-passwords.html

### Kibana

- **An index pattern for `logs-*` indices is automatically created.**

  It used to be required to manually create an index pattern for indices managed by Logstash, even when using the default Logstash indices. This is no longer the case since the output data is now being handled as a data stream, and Kibana automatically creates index patterns for these.

### Enterprise Search

- **Kibana is now the new management interface, and the only one available moving forward.**

  The old standalone Enterprise Search interface was removed in v8.0.0.

  Ref: https://www.elastic.co/guide/en/enterprise-search/current/user-interfaces.html


antoineco:
Note to self: Elasticsearch's API key service can't be enabled without TLS anymore, but the Enterprise Search extension [requires it](https://github.com/deviantony/docker-elk/tree/master/extensions/enterprise-search#enable-elasticsearchs-api-key-service).

Here is the corresponding Elasticsearch log message surfaced by the automated tests:

```json
{
  "@timestamp":"2021-01-16T21:56:54.454Z",
  "log.level": "WARN",
  "message":"HTTPS is required in order to use the API key service; please enable HTTPS using the [xpack.security.http.ssl.enabled] setting or disable the API key service using the [xpack.security.authc.api_key.enabled] setting",
  "service.name":"ES_ECS",
  "process.thread.name":"main",
  "log.logger":"org.elasticsearch.bootstrap.BootstrapChecks",
  "type":"server",
  "node.name":"30967a0e299e",
  "cluster.name":"docker-cluster"
}
```

This means we have to remove the Enterprise Search extension from the `master` branch, and keep it only on the `tls` branch.

PR #554, opened 2020-11-16
relaced single quota to double quota in docker-compose.yaml

gyulaweber:
When I wanted to start docker-compose the services, it returned a lots of errors and traces, I think the bottom line is the most relevant: 
`docker.errors.DockerException: Error while fetching server API version: '<' not supported between instances of 'Retry' and 'int'`
After I searched for this error message, found the problem and solution, which is to replace single quotes to double ones, and it solved the issue.

I experienced the same problem on Arch linux / docker compose 1.27.3 / docker 19.03.13-ce,
and debian buster/sid 1.17.1 / docker 19.03.13

I hope it helps

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '3.2'
+version: "3.2"
 
 services:
   elasticsearch:


antoineco:
Interesting, I'm wondering why this is happening since the text between ' ' and " " is interpreted the same in YAML in the absence of special characters.

Would you mind sharing the full error log and Compose file just so I can understand the actual reason for the error?

antoineco:
@gyulaweber you're not using Ansible, are you?

gyulaweber:
> @gyulaweber you're not using Ansible, are you?

I do, but not for this stack; I'll check it again and get back to you in a few days.

gyulaweber:
I've tried to reproduce this issue to get the full error message / environment details, but it seems I've upgraded docker-compose and other things in the meanwhile, so the problem is gone: if I change back to single quotes, it still works. Next time I'll collect all debugging information before opening a PR; sorry for the distraction.

antoineco:
OK thanks for the confirmation 👍 I'd also rather stick to single quotes if possible because double quotes can have a different meaning in YAML.

If the issue occurs again, feel free to reopen!

PR #564, opened 2021-01-12
Convert configuration files to v2 format

antoineco:
Convert all configuration files to `config_version: 2` (SG 7 format).


deviantony:
Interesting format update.

antoineco:
Yeah, it's a bit more explicit than the first version which doesn't hurt in terms of readability.

deviantony:
Allgood :+1:

Issue #550, opened 2020-11-07
Mention possible challenges with firewalld


antoineco:
A user reported issues running the stack due to `firewalld` blocking traffic on the `elk` bridge in #541. Another one reported similar problems in Gitter.

Similarly to what we did with SELinux, we should warn users about possible challenges using Compose on a machine running `firewalld`.

antoineco:
Dropping this great article here for future references: https://www.howtogeek.com/devops/how-to-use-docker-with-a-ufw-firewall/

PR #574, opened 2021-02-03
Add conntrack for fixing up remote addresses

### LINKED ISSUE #241: Logstash: Preserve Host IP
Hey,

I'm collecting some inputs via http plugin. I'd like to seperate data by the hosts IP address.  
  
Due to the fact the docker-elk service is creating his network named `dockerelk_elk` using the `bridge` driver, any connection is routed through the gateways of `dockerelk_elk` if it's not coming from within this network.   

When previewing the collected logs within Kibana, the `host` field always contains the gateways IP address because the requests came from the outside of `dockerelk_elk`. 

image

Does anyone has an idea how to preserve the IP address the request is coming from?

I found this issue from the docker repo <https://github.com/docker/libnetwork/issues/1994> suggesting to use the `host` network driver. So I started `elasticsearch` and `kibana` via the docker-compose service. Afterwards I started logstash via 

````
docker run -it --net=host -v ./logstash/pipeline/:/usr/share/logstash/pipeline/ -v ./logstash/config/logstash.yml:/usr/share/logstash/config/logstash.yml docker.elastic.co/logstash/logstash:6.1.3
````

In fact, from host network this docker container running logstash cannot connect to another network like `dockerelk_elk`. I ran `docker network connect dockerelk_elk [LOGSTASH_CONTAINER_ID]` just to get an error

````
Error response from daemon: container sharing network namespace with another container or host cannot be connected to any other network
````

So it seems to me I have to run all three containers on the same network to get this up and running. 

Or does anybody else have a suggestion?

Best,
huppys

orgads:
Fixes #241

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -69,6 +69,14 @@ services:
     depends_on:
       - elasticsearch
 
+  conntrack:
+    image: cap10morgan/conntrack
+    depends_on:
+      - logstash
+    network_mode: host
+    privileged: true
+    command: -D --proto udp
+
 networks:
   elk:
     driver: bridge


antoineco:
Thanks for proposing this change. I believe it is too opinionated to be included in the stack though, since it is related to the network plumbing on the host, and not to the Elastic stack per-se. I'm not convinced everybody wants us to interact with their host without permission, besides it involves running a random image from the internet in privilege mode.

orgads:
Good points. Thanks for the review :)

antoineco:
I think an entry in the docs would be **super useful** though, like a troubleshooting section towards the end where we include this, the note about SELinux and maybe a note about firewalls (#550).

You're more than welcome to tackle this, or I can give it a shot myself and assign you as a reviewer if you prefer.

orgads:
I don't know enough about these subjects to document them. I'll be happy to review it though :)

PR #597, opened 2021-06-23
Use built-in Compose v2

antoineco:
Marking as draft because the feature is currently in beta, and not yet included into the Docker client as of Docker 20.10.

---

Updating the CI test suite to use the new [Compose V2](https://docs.docker.com/compose/cli-command/), which is built into Docker as the `docker compose` command.


.github/workflows/scripts/run-tests-logspout.sh
```diff
@@ -23,7 +23,7 @@ poll_ready "$cid_ls" "http://${ip_ls}/health"
 #   2021/01/07 16:14:52 # logspout v3.2.13-custom by gliderlabs
 #
 # which we expect to find by querying:
-#   docker.image:"docker-elk_logspout" AND message:"logspout gliderlabs"~3
+#   docker.name:"/docker-elk_logspout_1" AND message:"logspout gliderlabs"~3
```
antoineco, line 26: Note: This change may seem unrelated, but it was necessary because the value of this field happens to be a hash with Compose v2. I'll confirm this is still the case once v2 is officially released.

.github/workflows/ci.yml
```diff
@@ -33,9 +35,19 @@ jobs:
...
 
+          # Enable support for Compose V2 (beta)
+          #
+          # Instructions:
+          # https://github.com/docker/compose-cli#compose-v2-aka-local-docker-compose
+
+          curl -sLO "$COMPOSE_V2_URL"
+          chmod +x docker-compose-linux-amd64
+          mkdir -p ~/.docker/cli-plugins
+          mv docker-compose-linux-amd64 ~/.docker/cli-plugins/docker-compose
```
antoineco, line 46: Note: Will be removed when v2 is officially released.

.github/workflows/scripts/run-tests-logspout.sh
antoineco: Related to https://github.com/docker/compose-cli/issues/1972

PR #623, opened 2021-09-16
Update to v7.14.1

Jordanlelay:


antoineco:
@Jordanlelay thanks for the contribution 🙌 

We usually don't push individual commits to branches other than `main`. Instead, we push all features and stack version updates to `main`, then rebase other branches onto it to keep the history "on top of main".

I just ran `git rebase main` on the `tls` branch and (force-)pushed it. You may need to perform a Git reset in order to cleanly get the changes locally:

```
git fetch origin
git checkout tls
git reset --hard origin/tls
```

antoineco:
The stack is now on v7.15.0, including the `tls` branch, so I'm going to consider this PR as obsolete.

PR #634, opened 2021-11-03
Use token from GitHub App in auto-update workflow

### LINKED ISSUE #552: Automated stack version updates do not trigger CI run
730ffa8de5f4f7c7c95d8a674744a0e09fe072ba introduced a new workflow that periodically checks the latest available release of the Elastic stack, then sends a PR if that version diverges from the one on the `main` branch.

Currently, those automated PRs do not trigger the CI workflow, so we don't know whether the changes break something until they land on `master`.

Since this is a known limitation when using the [GITHUB_TOKEN](https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#about-the-github_token-secret) and the workaround is already documented, I am opening this issue so that we can tackle it at a later point.

> If you want pull requests created by this action to trigger an `on: push` or `on: pull_request` workflow then you cannot use the default `GITHUB_TOKEN`. See the [documentation here](https://github.com/peter-evans/create-pull-request/blob/master/docs/concepts-guidelines.md#triggering-further-workflow-runs) for workarounds.
>
> -- https://github.com/peter-evans/create-pull-request#action-inputs

antoineco:
Per the instructions found at [Authenticating with GitHub App generated tokens](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens).

Requires the `docker-elk Updater` app referenced by `{{ secrets.APP_ID }}` to be first installed in the repo.

Example of PR generated by impersonating the app: https://github.com/antoineco/docker-elk/pull/19

Closes #552

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -82,10 +82,19 @@ jobs:
               echo '::set-output name=has_changes::true'
           fi
 
+      - name: Impersonate update bot
+        uses: tibdex/github-app-token@v1
+        id: generate-token
+        if: steps.update-files.outputs.has_changes
+        with:
+          app_id: ${{ secrets.APP_ID }}
+          private_key: ${{ secrets.APP_PRIVATE_KEY }}
+
       - name: Send pull request to update to new version
         if: steps.update-files.outputs.has_changes
         uses: peter-evans/create-pull-request@v3
         with:
+          token: ${{ steps.generate-token.outputs.token }}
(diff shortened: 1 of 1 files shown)


deviantony:
@antoineco should be good to go now :-) Nice piece of work!

antoineco:
Thanks for the quick answer! 👍

PR #662, opened 2022-01-21
Added a environmental variables to be pulled from the .env file

skibum1869:
The default username and password for boot are defined in the .env file for the first boot. 
The Kibana URL is set for server.publicBaseUrl: as an environmental variable in the .env file.


antoineco:
@skibum1869 if that change is really targeted at docker-elk and no at your own fork, I believe it contains a few undesired changes:

- It contains more changes than it claims it does.
- `elastic` is the only user enabled the first time you start Elasticsearch, so making it "configurable" that way is counter intuitive because any other value would prevent the stack from starting (unless someone manually created an alternative user). Besides, we encourage users **not** to use a single privileged user in our README. This PR breaks that principle (sort of).
- We also indicate in our README that the `ELASTIC_PASSWORD` is a **bootstrap** password that's only used on the **first boot**. After that first boot, you shouldn't ever set this variable again because it will be ignored. A lot of people will get bit by that if they expect the `.env` file to be a real source of truth.

skibum1869:
Your response brings up some concerns I had not thought of. I'm closing this PR.

Issue #666, opened 2022-02-11
Pre-configured user is forbidden


MartinJohns:
### Problem description

The Kibana container fails to start due to the pre-configured user "elastic"

```
[root] Error: [config validation of [elasticsearch].username]: value of "elastic" is forbidden. This is a superuser account that cannot write to system indices that Kibana needs to function. Use a service account token instead. Learn more: https://www.elastic.co/guide/en/elasticsearch/reference/8.0/service-accounts.html
    at ensureValidConfiguration (/usr/share/kibana/src/core/server/config/ensure_valid_configuration.js:25:11)
    at Server.preboot (/usr/share/kibana/src/core/server/server.js:160:5)
    at Root.preboot (/usr/share/kibana/src/core/server/root/index.js:48:14)
    at bootstrap (/usr/share/kibana/src/core/server/bootstrap.js:99:9)
    at Command.<anonymous> (/usr/share/kibana/src/cli/serve/serve.js:216:5)
```

### Extra information

Git Commit: [33a96e571a816b3417542496de16bbb2dfeb5afe](https://github.com/deviantony/docker-elk/commit/33a96e571a816b3417542496de16bbb2dfeb5afe)

This is most likely due to the recent update to 8.0.

#### Stack configuration

No changes. Just clone, build and start.

#### Docker setup

```console
$ docker version

Client:
 Cloud integration: v1.0.22
 Version:           20.10.12
 API version:       1.41
 Go version:        go1.16.12
 Git commit:        e91ed57
 Built:             Mon Dec 13 11:44:07 2021
 OS/Arch:           windows/amd64
 Context:           default
 Experimental:      true

Server: Docker Desktop 4.5.0 (74594)
 Engine:
  Version:          20.10.12
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.16.12
  Git commit:       459d0df
  Built:            Mon Dec 13 11:43:56 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.12
  GitCommit:        7b11cfaabd73bb80907dd23182b9347b4245eb5d
 runc:
  Version:          1.0.2
  GitCommit:        v1.0.2-0-g52b36a2
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

```console
$ docker-compose version

Docker Compose version v2.2.3
```

#### Container logs

```console
$ docker-compose logs

[...]

PR #671, opened 2022-02-14
Automate management of roles and user passwords required in v8.0.0

### LINKED ISSUE #673: Got error Kibana server is not ready yet.
i get the error Kibana server is not ready yet. I'm using v8, how do I fix it?
`[2022-02-16T03:02:58.666+00:00][INFO ][plugins.screenshotting.config] Chromium sandbox provides an additional layer of protection, and is supported for Linux Ubuntu 20.04 OS. Automatically enabling Chromium sandbox.
[2022-02-16T03:03:00.190+00:00][ERROR][elasticsearch-service] Unable to retrieve version information from Elasticsearch nodes. connect ECONNREFUSED 172.18.0.2:9200
[2022-02-16T03:03:01.090+00:00][INFO ][plugins.screenshotting.chromium] Browser executable: /usr/share/kibana/x-pack/plugins/screenshotting/chromium/headless_shell-linux_x64/headless_shell
[2022-02-16T03:03:04.149+00:00][ERROR][elasticsearch-service] Unable to retrieve version information from Elasticsearch nodes. security_exception: [security_exception] Reason: unable to authenticate user [kibana_system] for REST request [/_nodes?filter_path=nodes.*.version%2Cnodes.*.http.publish_address%2Cnodes.*.ip]
`

antoineco:
### Context

With the release of Elastic v8.0.0, users lost the ability to "clone and run" docker-elk due to a restriction on the usage of the bootstrap `elastic` user in Kibana. They now have to go through [manual initialization steps](https://github.com/deviantony/docker-elk/blob/641290c20af46e995fa6ac7c8a0584b96b17c8a5/README.md#initial-setup) in order to use docker-elk, namely:

1. run Elasticsearch in isolation (to set up users before starting other components)
1. create a password for the `kibana_system` user
1. write that password to the `kibana.yaml` file
1. start the rest of the stack

I feel that this will frustrate users (we already got 3 issues opened related to that topic) and somewhat defies the purpose of docker-elk.

### Proposal

**I think we can keep the setup easy without compromising on flexibility.**

In this PR, I drafted an approach that relies on a new one-off `setup` service injected into the main Compose file.
The `setup` service simply runs a Bash script that waits for Elasticsearch to become reachable, then sets up the users and roles required for the stack to run:

- `kibana_system` built-in user
- `logstash_internal` custom user with `logstash_writer` custom role

There are a few reasons why I think this is acceptable:

- (the obvious) This restores the "clone and run" experience I mentioned earlier. `git clone && docker-compose up` works again.
- The `setup` service isn't a hard dependency. If it runs, it ensures that things are set up properly. If not, the stack _will_ run anyway (but people need to perform the initial setup steps on their own).
- In case Kibana and/or Logstash start before their respective user has been set up, they simply retry for a while, so the experience is seamless and non-racy.
- If a password isn't set in the `.env` file, the script skips the management of the corresponding user. The `setup` service will run, do nothing, and exit with 0.
- It encourages best-practices. We used to connect Logstash to Elasticsearch using the `elastic` superuser. We can now inject a role tailored for Logstash instead (as described in the Logstash docs), and switch to an unprivileged user _by default_.

Here is a sample log of the `setup` service, printed on an initial run of the stack on v8.0.0:

```
docker-elk-setup-1  | [+] Waiting for availability of Elasticsearch
docker-elk-setup-1  |    ⠿ Elasticsearch is running
docker-elk-setup-1  | [+] Role 'logstash_writer'
docker-elk-setup-1  |    ⠿ Creating/updating
docker-elk-setup-1  | [+] User 'kibana_system'
docker-elk-setup-1  |    ⠿ User exists, setting password
docker-elk-setup-1  | [+] User 'logstash_internal'
docker-elk-setup-1  |    ⠿ User does not exist, creating
```

### Caveat

This approach is incompatible with our "Experimental support for Swarm", because there is no `docker-compose build` equivalent for Swarm, so no way to easily build the `setup` container.

Supporting Swarm comes with a lot of limitations anyway and there are better setups out there for Swarm/Kubernetes, so this might be a good excuse to say good bye to our `docker-stack.yml` file.

Related to #666
Closes #673

---

Marking as draft because this might need some internal discussions and updates to the docs.
Users are welcome to chime in! cc. @MartinJohns @kennyparsons


deviantony (approved):
> I feel that this will frustrate users (we already got 3 issues opened related to that topic) and somewhat defies the purpose of docker-elk.

Definitely agree with that one.

I unfortunately don't have the time to review this properly / test this at the moment but I agree on the following points:

* Being able to clone and run this project seamlessly
* The removal of Swarm support. This project was designed to provide a quick entrypoint to test the ELK stack and I feel like Docker is the best choice for that, Swarm less so (and Swarm is being less and less maintained anyway).

You got my approval👍

antoineco:
Thanks @deviantony 🙌 
I'll update the docs before merging.

kennyparsons:
Have you looked into docker's own support for init containers? 

My theory is that we can use an init container (a specially designed container, different than a standard container) to do this. 

But back to your solution: what about when a user brings the stack down and back up. Will the setup container re-run? Is there a pre-flight check to see if the passwords were already changed?

antoineco:
@kennyparsons thanks for your comments!

### Init container

An init container would execute something _before_ Elasticsearch starts (e.g. create a CA and some TLS certificates).

In our case, we need Elasticsearch to be **up** in order to set up our users and roles, so we can't rely on an init container unfortunately (although that would feel more elegant).

### Setup container

The `setup` one-off container runs on every `compose up`, unconditionally. I believe it's OK because updating a user or role is a no-op operation if that user or role is already in the expected state.

So basically, the `.env` file is the _source of truth_ when it comes to passwords. User credentials for Logstash and Kibana's users are always enforced from it, guaranteeing that good operations are restored upon a stack restart in case passwords "drifted".

In the eventuality where users _remove_ passwords completely from the `.env` file, but don't disable the `setup` container, password resets are simply [not performed](https://github.com/deviantony/docker-elk/blob/e6f0968fd6f83ca72c9a02efbd055044466e7a95/setup/entrypoint.sh#L61), and users are free to manage everything manually (like writing passwords in individual config files, or inside the Compose file instead of the `.env` file, etc.).

kennyparsons:
Ah, fair point on the init container needing elastic to be up. 

So essentially, the passwords are being set in the .env and being "changed" on every startup?

antoineco:
That's right.

Another reason for this is that there is no way to check the existing value of a user's password. Those are encrypted. We can only check whether a user exists, is enabled, etc. but not the value of their current password.

I'm wondering what kind of experience it would result in if we ensured that the `setup` container ran only _once_. Say, what is the expected result if a password is updated directly into Elasticsearch (e.g. via the Kibana UI) and the value from the `.env` file doesn't match anymore? Right now we "reconcile" from the `.env` file, but do you think that could be problematic?

deviantony:
@antoineco my two cents:

* Maybe we could only use the .env values as bootstrap values. After all, if a user changes the credentials of an account at runtime and it gets overridden after reboot/restart of the stack that would be a weird user experience IMO. I feel like not reconciling is more natural/expected from my perspective.

> Right now we "reconcile" from the .env file, but do you think that could be problematic?
As long as it is documented I don't think so.

* A simple solution to run the setup container only once would be to provide it with a volume, write an empty file in there after the first successful setup and check for the existence of that file when starting the container - if it exists, exit without running the bootstrap logic.

kennyparsons:
> * check for the existence of that file when starting the container

This, IMHO, is the right way to do it. But, I think your current PR is sufficient to get it moving in the right direction. But a "pre-flight" check on the setup container to see if the file exists first would preferred.

antoineco:
>> check for the existence of that file when starting the container
> 
> This, IMHO, is the right way to do it.

I'll add this then, and document accordingly. It is a minimal change and sounds relatively uncontroversial. "We set things up for you, then you're on your own."

mehmetkurt (approved):

Issue #677, opened 2022-02-26
new install cannot authenticate


mAineAc-:
### Problem description
I run docker compose everything completes. I check the logs to verify set up is complete. I can go to the ,env files and see the password for elastic is changeme and when I go to log in on the page it tells me the password is invalid. I have cleaned out all images for all docker sessions and started right from scratch and still get the same thing. I can go into the elastic config and change to no security and it brings it up so that i can do stuff but I don't see a way to create new users when I do this. I have tried changing the default password to other passwords to no avail. I still get an authentication failed when trying to log in. 

#### Stack configuration

maineac@elastiflow:~/docker/docker-elk$ git diff
diff --git a/elasticsearch/config/elasticsearch.yml b/elasticsearch/config/elasticsearch.yml
index 3e82379..f03613b 100644
--- a/elasticsearch/config/elasticsearch.yml
+++ b/elasticsearch/config/elasticsearch.yml
@@ -8,5 +8,5 @@ network.host: 0.0.0.0
 ## X-Pack settings
 ## see https://www.elastic.co/guide/en/elasticsearch/reference/current/security-settings.html
 #
-xpack.license.self_generated.type: trial
+xpack.license.self_generated.type: basic
 xpack.security.enabled: true
maineac@elastiflow:~/docker/docker-elk$

#### Docker setup

```console
$ docker version

maineac@elastiflow:~/docker/docker-elk$ sudo docker version
[sudo] password for maineac:
Client:
 Version:           20.10.7
 API version:       1.41
 Go version:        go1.13.8
 Git commit:        20.10.7-0ubuntu5~20.04.2
 Built:             Mon Nov  1 00:34:17 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server:
 Engine:
  Version:          20.10.7
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.8
  Git commit:       20.10.7-0ubuntu5~20.04.2
  Built:            Fri Oct 22 00:45:53 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.5.5-0ubuntu3~20.04.1
  GitCommit:
 runc:
  Version:          1.0.1-0ubuntu2~20.04.1
  GitCommit:
 docker-init:
  Version:          0.19.0
  GitCommit:

```

```console
$ docker-compose version

maineac@elastiflow:~/docker/docker-elk$ sudo docker-compose version
[...]

Issue #667, opened 2022-02-11
Kibana exits due to changed password in config


kennyparsons:
### Problem description

I followed the initial setup instructions to change passwords for elastic and kibana_system (as well as all other default users). But before that, the command to just bring up elasticsearch didn't work: 
```bash
docker-compose up elasticsearch -d
Creating network "docker-elk_elk" with driver "bridge"
ERROR: No such service: -d
```

So i pivoted and just did `docker-compose up elasticsearch`, which ran it in the foreground. No biggy. I ran the password commands in a separate window. I changed the passwords, edited the two config files mentioned, saved it, and moved on. So far so good.

Now i switch back to elasticsearch running in the foreground, kill it (ctrl-c), and then bring up the whole stack with `docker-compose up -d`.

I check the containers and kibana quits immediately. I checked the logs. The password generated started with an illegal character (`*`). So I put single quotes around it and it worked. Might be nice to have that defaulted to be in quotes, if passwords are randomly generated. 

### Extra information

The only remaining issue: `docker-compose up elasticsearch -d` didn't work. Any ideas why it does for you and not for me?

PR #678, opened 2022-03-01
Remove quotes from passwords

pleeplop:
In my environment default passwords don't work because single quotes are not removed.
So the password for Kibana web UI is `'changeme'` but it should be `changeme` as explained in the README

Environment:
Ubuntu 20.04.3 LTS
docker-compose version 1.25.0
Docker version 20.10.7

### DIFF
--- a/.env
+++ b/.env
@@ -7,16 +7,16 @@ ELASTIC_VERSION=8.0.0
 #
 # Superuser role, full access to cluster management and data indices.
 # https://www.elastic.co/guide/en/elasticsearch/reference/current/built-in-users.html
-ELASTIC_PASSWORD='changeme'
+ELASTIC_PASSWORD=changeme
 
 # User 'logstash_internal' (custom)
 #
 # The user Logstash uses to connect and send data to Elasticsearch.
 # https://www.elastic.co/guide/en/logstash/current/ls-security.html
-LOGSTASH_INTERNAL_PASSWORD='changeme'
+LOGSTASH_INTERNAL_PASSWORD=changeme
 
 # User 'kibana_system' (built-in)
 #
 # The user Kibana uses to connect and communicate with Elasticsearch.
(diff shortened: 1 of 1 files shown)


antoineco:
@pleeplop thanks for the suggestion 🙌 

That is a **known issue** with Compose versions below `v1.26.0`. For this reason, we [only support Compose `v1.26.0` and above](https://github.com/deviantony/docker-elk/blob/159c934ad28a62a4022beab23ce4fb98d5c5c98f/README.md#host-setup).

Extract from the README:

> **Host setup**
> - [Docker Compose](https://docs.docker.com/compose/install/) version **1.26.0** or newer

Versions between `v1.22.0` and `v1.25.5` are known to work as well, but require omitting quotes like you suggested:

```ini
# .env
ELASTIC_VERSION=8.0.0
ELASTIC_PASSWORD=changeme
LOGSTASH_INTERNAL_PASSWORD=changeme
KIBANA_SYSTEM_PASSWORD=changeme
```

I have to reject this PR because we **don't want to remove quotes from the default `.env` file** for the following reasons:

1. `bin/elasticsearch-reset-password` generates passwords with **special characters** (`$#?!=`...), which makes quoting necessary in some cases (see #667).
     
    For example, an `.env` file with unquoted values containing special characters would cause issues with tools that read `.env` files using a shell (such as (`direnv`)[https://direnv.net/])

1. This is only an issue with _unsupported_ versions of Compose.

    `v1.25.0` is two years old and we can't possibly keep supporting old versions forever. The issue with `.env` files is one reason for this. Bumping the [format version](https://docs.docker.com/compose/compose-file/compose-file-v3/) of Compose files is another one.

A few references:
- https://github.com/docker/compose/issues/2854
- https://docs.docker.com/compose/release-notes/#1260
- #677

pleeplop:
It makes sense.
Thank you for the feedback.

PR #683, opened 2022-03-07
Ogs 407 create carrier order fix readiness check

KaungMyatChanThar:
Ogs 407 create carrier order fix readiness check


antoineco:
(Assuming wrong base repo and closing)

PR #699, opened 2022-04-12
restart policy for contianer

saeed-mcu:


antoineco:
Closing because apparently sent to the wrong base repo.

PR #725, opened 2022-06-18
feat: healthcheck for elasticsearch and kibana

117503445:
Referring to [get-started-stack-docker](https://www.elastic.co/guide/en/elastic-stack-get-started/current/get-started-stack-docker.html#docker-compose-file), added healthcheck for elasticsearch and kibana, and ensured that kibana starts after elasticsearch

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,6 +23,7 @@ services:
       KIBANA_SYSTEM_PASSWORD: ${KIBANA_SYSTEM_PASSWORD:-}
     networks:
       - elk
+      
 
   elasticsearch:
     build:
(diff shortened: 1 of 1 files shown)


antoineco:
@117503445 thanks for the contribution!

Health checks aren't necessary here because Elastic components are already resilient and wait for Elasticsearch to be ready before moving on.

I think we are better off relying on that built-in resilience instead of hacking things like grepping the output of curl. That approach might even be inaccurate because a 200 response from Elasticsearch isn't an indicator that all Elasticsearch APIs are ready, it just indicates that it has started.

PR #746, opened 2022-08-03
Update .env

igerlster:
Quotes are used as password, so login was set to 'changeme' and not changeme.

### DIFF
--- a/.env
+++ b/.env
@@ -7,16 +7,16 @@ ELASTIC_VERSION=8.3.3
 #
 # Superuser role, full access to cluster management and data indices.
 # https://www.elastic.co/guide/en/elasticsearch/reference/current/built-in-users.html
-ELASTIC_PASSWORD='changeme'
+ELASTIC_PASSWORD=changeme
 
 # User 'logstash_internal' (custom)
 #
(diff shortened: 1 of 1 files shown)


antoineco:
That is because you are using a version of Compose we don't support, as mentioned in the warning inside the [**Host Setup** section of the README](https://github.com/deviantony/docker-elk#host-setup).

Summary:
- Versions of Compose below 1.26 have a [known bug](https://github.com/docker/compose/issues/2854) that make them interpret quotes **literally**.
- We need to be conservative with our usage of quotes because versions _above_ 1.26 _do require quotes_ when values contain special characters — such as `$&!*...` — which might be interpreted. (So do other popular Unix tools that read `.env` files, such as [direnv](https://direnv.net/)).
    Those characters are common in passwords, including the ones generated by Elasticsearch's password tool, which usage we recommend in the README.

As you can see, we either support a version of Compose that is over two years old, or we quote passwords to prevent our users from running into issues such as in #667.
If I have to decide between the two, I choose to support newer versions anytime.

igerlster:
Oh, Sorry. We did miss this. Thanks for the reply

PR #811, opened 2022-12-23
Use restart policy unless-stopped

h44z:
Automatically start elk stack after a reboot

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -57,6 +57,7 @@ services:
       discovery.type: single-node
     networks:
       - elk
+    restart: unless-stopped
 
   logstash:
     build:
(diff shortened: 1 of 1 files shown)


antoineco:
@h44z can you explain the rationale for it?

I'm not against merging this if it solves an actual problem but, as far as I know, all Elastic components are resilient to failures and continue operating on (recoverable) errors.

When they actually exit, it's almost always due to unrecoverable errors: invalid configuration, insufficient file permissions, ... In which case the last thing you want is to restart automatically, because the problem needs to be addressed.

h44z:
Currently the docker-elk containers do not restart if the host machine (or the docker daemon) gets restarted. 
Thats why a [restart-policy](https://docs.docker.com/config/containers/start-containers-automatically/) makes sense I guess

antoineco:
Ah I didn't know this also applied to host restarts. Thanks for the pointer!

antoineco (approved):

Issue #839, opened 2023-03-25
Error with test suite workflow at "Execute core test suite" with clean template clone


AnonymousSmurf:
I am opening this issue because I still believe it is relevant, even though it may not be directly related to the stack itself.

### Problem description
I have created a repository using this template, but the CI workflows appear to be failing. Specifically, they all fail at the `Execute core test suite` step. To provide additional information, I have added some screenshots in the "Extra Information" section. However, I must admit that I am not yet very experienced with this code stack, so I apologize if I have overlooked a simple solution.

### Extra information
<img width="1422" alt="image" src="https://user-images.githubusercontent.com/45464235/227715976-5e82720c-8759-40fa-841e-a6136a4573b7.png">
<img width="1035" alt="image" src="https://user-images.githubusercontent.com/45464235/227715999-bb901716-6d0c-43c5-8683-3ba4f3d3f1ae.png">

#### Stack configuration
I am using version 8.6.2, which should be the latest version as far as I know. I will leave these fields blank because all configurations and versions are set to default, and it seems that the issues are related more to the workflows of the stack than the stack itself.

antoineco:
This is the faulty line:

https://github.com/deviantony/docker-elk/blob/b43620aee4146fbf947c44c9280cad9eb50e9b41/.github/workflows/scripts/lib/testing.sh#L74

If you clone the repository in a directory that is not named "docker-elk", or if you rename the "elk" network inside the Compose file, you also need to adjust the generated network name inside this command.

PR #844, opened 2023-03-28
Introduced -p flag to CI to prevent testsuit fails

AnonymousSmurf:
First and foremost, credit goes to 🎉 @antoineco 🎉 for coming up with the solution to [the issue](https://github.com/deviantony/docker-elk/issues/839) I was facing. 

## Issue

The problem that this pull request seeks to address is related to the CI workflow, which throws errors when the repository is not named docker-elk.

The issue is rooted in four files, which expect the containers to be called docker-elk. However, this naming convention is determined by the repository name, which can cause problems when it is different from docker-elk.

https://github.com/deviantony/docker-elk/blob/b43620aee4146fbf947c44c9280cad9eb50e9b41/.github/workflows/scripts/lib/testing.sh#L74

https://github.com/deviantony/docker-elk/blob/4ad25cf7292fab91fc27ad0064dbc737e6ac3cc6/.github/workflows/scripts/run-tests-logspout.sh#L45

https://github.com/deviantony/docker-elk/blob/4ad25cf7292fab91fc27ad0064dbc737e6ac3cc6/.github/workflows/scripts/run-tests-filebeat.sh#L40

https://github.com/deviantony/docker-elk/blob/4ad25cf7292fab91fc27ad0064dbc737e6ac3cc6/.github/workflows/scripts/run-tests-metricbeat.sh#L41

## Solution

To address this issue, I implemented changes to the CI workflow, which involved adding a "-p flag" to all docker-compose commands. This pull request introduces those changes, and I am pleased to report that I cloned the repository and ran the updated workflow successfully. Despite having a different repository name, all tests passed, as evidenced by the screenshot below.

image

P.S. I am still a novice at this, and any feedback would be greatly appreciated.

### DIFF
--- a/.github/workflows/spam-issue-close.yml
+++ b/.github/workflows/spam-issue-close.yml
@@ -17,7 +17,7 @@ jobs:
     steps:
       - name: Close
         id: close
-        uses: actions/stale@v7.0.0
+        uses: actions/stale@v8.0.0
         with:
           days-before-issue-stale: -1
           days-before-issue-close: 0


antoineco:
The approach looks good to me 👍 

There is an even simpler way: set the project name globally, once, using an environment variable:

- https://docs.docker.com/compose/environment-variables/envvars/#compose_project_name
- https://docs.github.com/en/actions/learn-github-actions/variables#defining-environment-variables-for-a-single-workflow

AnonymousSmurf:
I have created a global variable called `PROJECT_NAME`. The test suite has successfully passed without any issues. I think this pr can now be merged?

AnonymousSmurf:
No need to apologize. I have added the `COMPOSE_PROJECT_NAME` variable to the `.env` file and the testsuite is still passing. Therefore, I believe that the pull request can now be closed.

antoineco:
I would prefer if we kept that variable in the CI file like you did originally, since the change isn't relevant outside of the automated tests.

AnonymousSmurf:
Sure! I have added the changes back in. : )

antoineco:
Now the environment variable is `PROJECT_NAME` instead of `COMPOSE_PROJECT_NAME`, and the `-p` flags are back, but not needed.

For this feature, we need nothing but a single, global `COMPOSE_PROJECT_NAME` environment variable.

AnonymousSmurf:
I apologize for the misunderstanding. I have now added it as an environment variable inside the CI workflow. However, I seem to have accidentally closed the PR...

PR #859, opened 2023-05-09
Added Support for Let's Encrypt Certificates

mhmdksh:
Added the support for running the ELK Cluster with Let's Encrypt using Caddy to automatically generate the Certificates and configuring it as a reverse proxy. Based on a configurable .ENV value for the URL
I also updated the README.md to reflect those changes and provide guidance on how to use it


antoineco:
Thanks for your contribution! 🙌 

I love the idea of having better instructions for enabling TLS in Kibana than [just a link to the official docs](https://github.com/deviantony/docker-elk/blob/295ee80342984f025b41f43232a43ce987cdc1dc/kibana/config/kibana.yml#L27-L32). 

Introducing a full fledged external web server just for the side effect of having automatically issued certificates feels slightly against the [project philosophy](https://github.com/deviantony/docker-elk#philosophy) to me though. In my opinion, the changes suggested here are a case for _"use this repository as a template, tweak it, and make it your own"_.

My reasoning is, in no particular order:

- Kibana can serve requests _natively_ over HTTPS already. It doesn't technically need to be fronted by another web server.
- The choice for Let's Encrypt is opinionated. Popular, but opinionated. docker-elk strives to stay as neutral as possible when it comes to things which aren't part of the Elastic stack.
- Caddy would now become a non-optional component on the `tls` branch, and because it requires extra steps to be set up, docker-elk no longer works "out-of-the-box". Every new user will need to perform those extra steps to allow components to even start.
- Miscellaneous things, such as requiring a public URL, `tls_insecure_skip_verify` enabled, Caddy exposing Elasticsearch when the feature is for Kibana, etc.
- (Additionally, this PR introduces a lot of changes which aren't related to the suggested feature, such as README reformating and changes to container names.)

Instead, I think this content could be well suited for an additional page in the wiki related to enabling TLS in Kibana. This page could have for example two sections: a) request certificates only, with certbot; b) serve HTTPS over Caddy, with the instructions from this PR.

PR #873, opened 2023-06-13
Add Kibana health-check to avoid Fleet's "service_token" request errors. 

nuno-andre:
When deploying Fleet along the base stack, the fleet server tries to get a `service_token` from Kibana immediately, which slows down the deployment and makes a very noisy log with multiple `fleet-server` errors and retries. Adding this health-check, `fleet-server` will wait until Kibana has reached the `service_healthy` status.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -104,6 +104,12 @@ services:
       - elk
     depends_on:
       - elasticsearch
+    healthcheck:
+      test: curl -f http://localhost:5601 || exit 1
+      interval: 10s
+      timeout: 10s
+      retries: 10
+      start_period: 10s
     restart: unless-stopped
 
 networks:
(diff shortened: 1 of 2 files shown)


antoineco:
Thanks for your contribution 🙌 
I get your point, but I'm not convinced about the usefulness of such addition. In fact, with the flexibility that people have with regards to authentication, TLS, network, etc. adding those Docker health checks introduces more chances for errors than solutions.
If people change their configuration, Kibana might be up, running, and healthy, but dependencies might now stay down just because curl fails. Debugging such issues isn't obvious and, in a project that popular, I'm not willing to risk the friction without a good reason.

All Elastic components are already **built for resilience**. If a dependency is offline, the connection will simply be retried. You might find the messages "very noisy" and I get it, nobody likes to see errors, but they are absolutely harmless, and a sign that resilience works as expected. If it isn't broken, what are we even fixing?

Also letting Fleet fail and retry does not "slow down the deployment", not more than delaying its startup with curl requests.

PR #1041, opened 2024-12-19
changed the docker-compose local config directories.

stallieman:


stallieman (commented):
Looks like it works.

antoineco:
This was probably not meant to be opened upstream, so closing.

PR #1183, opened 2026-07-13
Sync from original repo

JeremyRms:


JeremyRms:
@antoineco my bad, wrong source

PR #1185, opened 2026-07-14
Filebeat added

jonmol:
Added filebeat to make it easier to both get data into Elasticsearch, but also so it's possible to use the project as a complete ingest pipeline test.

Made the following changes to logstash:

- Added `--config.reload.automatic` so filter changes are reloaded without a restart
- Changed logging to debug and added log4j filters so any problems parsing input data is easy to see
- Added a patterns directory for regex used by grok

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -88,9 +88,13 @@ services:
       context: logstash/
       args:
         ELASTIC_VERSION: ${ELASTIC_VERSION}
+    command:
+      - "--config.reload.automatic"
     volumes:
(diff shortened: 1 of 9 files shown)


antoineco:
Thanks for your contribution. We will not accept it because [extensions/filebeat](https://github.com/deviantony/docker-elk/tree/9.2511.1/extensions/filebeat) already exists, but if somebody looks for alternative examples they should be able to find your PR.

jonmol:
Makes complete sense, I had completely missed the extensions folder in the project. Sorry for wasting your time

